PR #20, opened 2026-01-25
feat: stream background Task sub-tools in real-time (WIP)

BowgartField:
## Summary

- Implement real-time streaming display of sub-tools when Tasks run with `run_in_background: true`
- Display Task prompt/instructions in the UI when expanding a Task

### Background Task Tailing System
- Created `src-tauri/src/chat/background.rs` module with:
  - `BackgroundTailerHandle` and `BackgroundTailers` types for managing tailers
  - `parse_background_task_result()` to extract `output_file` path from Task results
  - `spawn_background_tailer()` to tail background Task output files and emit `chat:tool_use` events with proper `parent_tool_use_id`
  - Proper cleanup when agents complete (on "result" message)

### Integration in Main Chat Flow
- Track pending background Tasks in `claude.rs` (detect `run_in_background: true` in Task tool_use)
- Spawn dedicated tailer for each background Task when its tool_result arrives
- Wait for all background tailers to complete before emitting `chat:done` (keeps status indicator active)
- Fixed content extraction bug: `tool_result.content` is an array, not a string

### UI Improvements
- Display Task prompt/instructions when expanding a Task in `ToolCallInline.tsx`

## What's NOT Working

Background Tasks are **interrupted** when the main Claude CLI process exits. The subagent output files show `[Request interrupted by user]` immediately after the main turn completes.

**Root cause:** Jean uses `--print` flag which causes Claude CLI to exit after each response. When the CLI exits, it sends interrupt signals to all running background agents.

## Solutions Under Consideration

| Option | Description | Pros | Cons |
|--------|-------------|------|------|
| **Keep CLI alive with dummy message** | Send a follow-up message to keep CLI running while Tasks execute | Localized change | Hacky, consumes tokens, adds visible message |
| **Use `--continue` instead of `--print`** | Interactive mode keeps CLI alive | Clean long-term solution | Major architectural change, high regression risk |
| **Resume interrupted Tasks** | Detect interruption, relaunch with `--resume <agentId>` | No change to main flow, Tasks resume where they stopped | Slight latency, need to track agentIds |
| **Accept limitation** | Document that background Tasks don't work in Jean | No code change | Feature doesn't work |

**Recommended:** Option 3 (Resume interrupted Tasks) seems cleanest - matches what a user would do manually.

## Test Plan

- [ ] Launch a chat that triggers background Tasks
- [ ] Verify sub-tools appear in real-time under the Task
- [ ] Verify Task prompt is visible when expanded
- [ ] Verify status indicator stays active while Tasks run
- [ ] Verify Tasks complete successfully (currently blocked by interruption issue)

### DIFF
--- a/src-tauri/src/chat/background.rs
+++ b/src-tauri/src/chat/background.rs
@@ -1,0 +1,401 @@
+//! Background Task tailing for sub-agent tool calls
+//!
+//! When a Task tool runs with `run_in_background: true`, its sub-tools are written
+//! to a separate output file. This module provides functionality to tail that file
+//! and emit tool events with the proper `parent_tool_use_id` for UI grouping.
+
+use std::collections::HashMap;
+use std::path::PathBuf;
+use std::sync::mpsc::{self, Sender};
+use std::sync::{Arc, Mutex};
+use std::thread;
+use std::time::Duration;
+
+use serde_json::Value;
+use tauri::{AppHandle, Emitter};
+
+use super::tail::{NdjsonTailer, POLL_INTERVAL};
+
+/// Payload for tool use events sent to frontend (matches claude.rs)
+#[derive(serde::Serialize, Clone)]
+struct ToolUseEvent {
+    session_id: String,
+    worktree_id: String,
+    id: String,
+    name: String,
+    input: serde_json::Value,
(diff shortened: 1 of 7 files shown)


andrasbacsai:
I solved this in a different way, but thanks for the PR! :)

PR #21, opened 2026-01-26
fix: show permission prompts immediately during streaming

Frenvius:
Remove the !isSending condition that blocked permission dialogs from appearing while Claude was still streaming.

Permission prompts (for Read, Grep, Bash, etc.) now appear immediately when Claude requests them, instead of waiting for the response stream to complete.

This matches Claude Code CLI behavior where permission prompts appear immediately when requested.


andrasbacsai:
Thank you for the PR, but this is a bit more complex than this. I tested it and it does now work properly.

PR #65, opened 2026-02-11
fix: Window control buttons not working

### LINKED ISSUE #59: Window control buttons (minimize, fullscreen, close) not working
On Ubuntu 24.04 Desktop, the window control buttons — minimize, fullscreen, and close — do not respond when clicked.
However, double-clicking on the top bar correctly toggles the window between maximized and windowed mode.

OS :  Ubuntu 24.04 Desktop

alepouna:
Fixes #59 

There are currently some issues where pressing X does not close the app rather just closes the wizard. 

To keep the scope of this PR small, it can be fixed in another PR.

Additionally, there are some quirks with how the interface is managed right now, like the settings windows shows up twice, etc. 

This was tested on macOS and Windows.

### DIFF
--- a/src/lib/commands/index.ts
+++ b/src/lib/commands/index.ts
@@ -5,6 +5,7 @@ import { appearanceCommands } from './appearance-commands'
 import { notificationCommands } from './notification-commands'
 import { projectCommands } from './project-commands'
 import { githubCommands } from './github-commands'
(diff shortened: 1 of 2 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #78, opened 2026-02-13
feat: add Jean Client app for remote server connection

yigitkonur:
## Summary

Adds a standalone **Jean Client** — a lightweight Tauri desktop app that connects to a remote Jean server over WebSocket. All domain operations (chat, projects, git) proxy through WebSocket to the server, while native Tauri features (window controls, zoom, clipboard) work locally.

**Key changes:**

- **`client-app/`** — New minimal Tauri shell (~100 lines of Rust) with `load_client_config` / `save_client_config` commands, proper capabilities, and relaxed CSP for remote connections
- **`ConnectionScreen`** — Login UI with server URL input and optional access token (auth is not required if the server has it disabled)
- **`ConnectionPane`** — Settings pane for managing/disconnecting the active server connection
- **`environment.ts`** — New `isClientApp()` and `isServerApp()` detection functions. `isClientApp()` uses both Vite env var and Tauri initialization script (`__JEAN_CLIENT_MODE__`) for reliable detection
- **`transport.ts`** — Three-mode routing: Server app → Tauri IPC; Client app → WebSocket (with local IPC allowlist for config commands); Browser → WebSocket. Configurable WS URL from localStorage
- **9 files audited** — Changed `isNativeApp()` → `isServerApp()` for features requiring a local backend (terminal, file picker, editor/finder launch, CLI management)
- **PreferencesDialog** — Server-only panes hidden in client mode; new "Connection" pane shown only in client mode

**Architecture:**
```
isNativeApp()  → true in ANY Tauri shell (for native window features)
isClientApp()  → true in Jean Client only (thin remote client)
isServerApp()  → true in full Jean app only (local backend features)
```

## Test plan

- [ ] Build main Jean app (`bun run tauri:dev`) — verify no regressions in browser mode and native mode
- [ ] Build client app (`cd client-app && bun run tauri:dev`) — verify ConnectionScreen appears
- [ ] Connect to a Jean server (with and without auth token) — verify WebSocket connection establishes
- [ ] Verify server-only features are hidden in client mode (no terminal panel, no "Open in Finder/Editor", no CLI settings, no Web Access pane)
- [ ] Verify native features work in client mode (zoom, traffic lights, clipboard, keyboard shortcuts)
- [ ] Verify "Connection" settings pane appears in Preferences in client mode
- [ ] Production build: `cd client-app && bun run tauri:build:macos`

### DIFF
--- a/client-app/src-tauri/src/lib.rs
+++ b/client-app/src-tauri/src/lib.rs
@@ -64,6 +64,47 @@ fn save_client_config(app: tauri::AppHandle, config: ClientConfig) -> Result<(),
 }
 
 pub fn run() {
+    // FIX: Avoid WebKit GBM buffer errors on Linux (especially NVIDIA)
+    //
+    // This issue occurs when using transparent windows with WebKitGTK on Linux,
+    // particularly with NVIDIA GPUs. The error "Failed to create GBM buffer of size NxN: Invalid argument"
+    // is caused by incompatibilities between hardware-accelerated compositing and certain
+    // GPU drivers/compositors.
+    //
+    // Related issues:
+    // - https://github.com/tauri-apps/tauri/issues/13493
+    // - https://github.com/tauri-apps/tauri/issues/8254
+    // - https://bugs.webkit.org/show_bug.cgi?id=165246
+    // - https://github.com/tauri-apps/tauri/issues/9394 (NVIDIA problems doc)
+    //
+    // The fix disables problematic GPU compositing modes. Users can override via env vars:
+    // - JEAN_FORCE_X11=1 to force X11 backend (default: no)
+    // - WEBKIT_DISABLE_COMPOSITING_MODE=0 to re-enable GPU compositing (risky)
+    #[cfg(target_os = "linux")]
+    {
+        log::trace!("Setting WebKit compatibility fixes for Linux (client app)");
+
+        // Disable problematic GPU compositing modes
+        if std::env::var_os("WEBKIT_DISABLE_COMPOSITING_MODE").is_none() {
+            std::env::set_var("WEBKIT_DISABLE_COMPOSITING_MODE", "1");
+            log::trace!("WEBKIT_DISABLE_COMPOSITING_MODE=1");
+        }
+
+        // Disable DMABUF renderer (common cause of GBM errors)
+        if std::env::var_os("WEBKIT_DISABLE_DMABUF_RENDERER").is_none() {
+            std::env::set_var("WEBKIT_DISABLE_DMABUF_RENDERER", "1");
+            log::trace!("WEBKIT_DISABLE_DMABUF_RENDERER=1");
+        }
+
+        // Force X11 backend if Wayland causes issues
+        // Check if user explicitly wants X11 via environment variable
+        let force_x11 = std::env::var("JEAN_FORCE_X11").unwrap_or_else(|_| "0".to_string()) == "1";
+        if force_x11 && std::env::var_os("GDK_BACKEND").is_none() {
+            std::env::set_var("GDK_BACKEND", "x11");
+            log::trace!("GDK_BACKEND=x11 (forced by JEAN_FORCE_X11)");
+        }
+    }
+
     tauri::Builder::default()
         .plugin(tauri_plugin_log::Builder::new().build())
         .plugin(tauri_plugin_process::init())
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -98,6 +98,18 @@ function App() {
     if (!isClientApp()) return true // Not client mode — skip gate
     return !!localStorage.getItem('jean-client-server-url')
   })
+  const wsAuthError = useWsAuthError()
+
+  // Client mode: if auth becomes invalid and the token was cleared, return to setup.
+  useEffect(() => {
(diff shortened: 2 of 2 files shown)


yigitkonur:
Did a second pass over the whole PR. Went deeper this time — read every file line by line, traced call paths, and only touched things that are actually broken (not theoretical).

**10 files changed, +164 / −185 lines** (removed more code than added).

---

### Transport layer (`transport.ts`)

- **Deleted `getPreloadedData()`** — this function could never work. It called `.then()` on a resolved promise expecting a synchronous result, but `.then()` always runs as a microtask. The `return result` line always returned `null`. Zero callers in the codebase anyway, so just removed it.

- **Fixed the zombie WebSocket bug** — if you called `disconnect()` while `validateAndConnect()` had an in-flight fetch, the fetch would eventually resolve and silently open a new WebSocket that nobody asked for. Added an `AbortController` that `disconnect()` aborts, and the fetch catches `AbortError` and bails.

- **Added 15s timeout to the reconnect fetch** — the first-pass fix added a timeout to ConnectionScreen's fetch, but the transport layer's own auto-reconnect path (`validateAndConnect`) still had no timeout. A hanging server would freeze the reconnect loop indefinitely. Same AbortController approach now.

- **Added jitter to exponential backoff** — the delay was pure `2^n` with no randomization. If a server restarts and 50 clients are connected, they'd all retry at exactly 1s, 2s, 4s, etc. Added `base * (0.5 + random * 0.5)` to spread them out.

- **URL scheme validation in `getServerOrigin()`** — reads raw localStorage, which could theoretically be tampered via devtools/XSS. Now rejects anything that isn't `http://` or `https://`.

### Listener leaks (`pr-status.ts`, `useMainWindowEventListeners.ts`)

- **Fixed the PR status event listener leak** — `usePrStatusEvents` was pushing unlisten functions into an array via `Promise.all().then()`, but the cleanup function tried to iterate that array synchronously on unmount. The array was always empty at cleanup time. Every WS reconnect leaked another `pr:status-update` handler. Rewrote to use the `.then(f => f())` pattern that `useStreamingEvents` already does correctly.

- **Fixed the same bug in menu listeners** — `useMainWindowEventListeners` had the same race: `setupMenuListeners()` returns a promise, but cleanup read the result array synchronously. Now chains cleanup on the promise directly.

### Tauri security config (`tauri.conf.json`, `capabilities/default.json`, `Cargo.toml`)

- **Removed `unsafe-inline` from CSP `script-src`** — Vite production builds don't need it (no inline scripts in the output). With `unsafe-inline`, any HTML injection would become script execution, which combined with the `load_client_config` IPC command means full token exfiltration. Now `script-src 'self'` only.

- **Disabled asset protocol** — was scoped to `$APPDATA/**` which includes `client-config.json` (contains the server token). The client app doesn't use `asset://` for anything, so just turned it off.

- **Removed `process:allow-exit`, `process:allow-restart`** — nothing in the client frontend calls these. They just expanded the IPC attack surface for no reason.

- **Removed `core:event:allow-emit`, `core:event:allow-emit-to`** — the client communicates over WebSocket, not Tauri IPC events. No frontend code emits events. Removed.

### Connection UX (`ConnectionScreen.tsx`)

- **Added URL format validation** — previously, garbage input like `http://` or random text got auto-prefixed with `http://` and sent to fetch, which would hang for 15 seconds before timing out. Now validates with `new URL()` and shows "Invalid URL format" immediately.

- **Added a Cancel button** — when connecting, you had to wait the full 15 seconds if the server was slow. Now there's a Cancel button that aborts the fetch.

- **SSL error detection** — SSL/certificate errors were lumped into the generic "Could not reach the server" message. Now checks the error message for cert-related keywords and shows a specific hint about HTTPS misconfiguration.

### Rust / build (`lib.rs`, `Cargo.toml`)

- **Added Linux GPU workaround** — the main app already handles the WebKit GBM buffer crash on Linux+NVIDIA, but the client app didn't. Copied the same `WEBKIT_DISABLE_COMPOSITING_MODE` / `WEBKIT_DISABLE_DMABUF_RENDERER` env var approach, gated behind `#[cfg(target_os = "linux")]`.

- **Removed unused `dirs` crate** — listed in Cargo.toml but never imported. The app uses Tauri's built-in `app.path().app_data_dir()` instead.

---

### Investigated but intentionally left alone

- **No max-retry cap** — infinite retry at 30s intervals is the right behavior for a desktop app reconnecting to its server.
- **WS message validation** — already has type/id/event field checks in the onmessage handler, contrary to what the audit flagged.
- **`null as T` cast** — only in the E2E mock path where null is the intended return value.
- **Queue size cap** — the UI shows ConnectionScreen when disconnected, blocking any command invocations. Queue can't grow.
- **`window.location.reload()` on disconnect** — replacing this with graceful state reset would touch 3+ stores and the query cache. Reload is the pragmatic choice for v0.1.0.
- **ConnectionPane stale URL** — component mounts once and stays until disconnect (which triggers reload). URL can't diverge.
- **Shared `dist/` directory** — both Tauri configs point to the same `dist/`. Not great but rarely an issue in practice since you build one or the other.
- **Unix-only build commands** — project-wide pattern, not specific to this PR.
- **`opener:allow-open-url` scope** — actively used by GitHub integration (28+ call sites). Tauri v2 doesn't support URL scheme filtering in capabilities anyway.

PR #80, opened 2026-02-13
fix: prevent tokio thread pool starvation at 20+ concurrent sessions

yigitkonur:
## Problem

When running 20+ Claude sessions simultaneously, new sessions hang permanently on "Loading..." and existing simple data queries (preferences, session lists, etc.) stop responding entirely.

**Root cause:** `send_chat_message` is an `async fn` (runs on the tokio runtime), but it calls `execute_claude_detached` — a **synchronous blocking function** that internally runs `tail_claude_output`, which loops with `std::thread::sleep(50ms)` for the entire duration of a Claude response (often minutes).

Tokio sizes its thread pool to the number of logical CPUs. On an M4 Max (10 logical CPUs = 10 tokio worker threads), once ~10 sessions are actively streaming, every single tokio worker is blocked inside `thread::sleep`. No worker threads remain to dispatch any Tauri command — the app freezes.

**Reproduction:**
1. Open 15+ sessions in a project
2. Send messages to all of them
3. Try opening a new session or navigating — the UI hangs on "Loading..."

## Fix

Move the blocking `execute_claude_detached` call onto a dedicated OS thread using `std::thread::spawn` + `tokio::sync::oneshot`, so it no longer occupies tokio worker threads.

**Only file changed:** `src-tauri/src/chat/commands.rs` (~90 lines changed, same logic)

This follows the existing pattern already used in `naming.rs:764` (`spawn_naming_task`) for the same reason — offloading blocking work from the async runtime.

### Before

```rust
// Runs directly on a tokio worker thread — blocks it for minutes
let (pid, claude_response) = loop {
    match execute_claude_detached(&app, ...) { ... }
};
```

### After

```rust
// Offloaded to a dedicated OS thread — tokio workers stay free
let (tx, rx) = tokio::sync::oneshot::channel();
std::thread::spawn(move || {
    let result = loop {
        match execute_claude_detached(&bl_app, ...) { ... }
    };
    let _ = tx.send(result);
});
let (pid, claude_response) = rx.await.map_err(|_| "thread panicked")??;
```

## Why this is safe

| Concern | Status | Reasoning |
|---------|--------|-----------|
| `AppHandle` thread safety | Safe | Tauri's `AppHandle` is `Send + Sync` by design; `emit_all()` works from any thread |
| `with_sessions_mut` (retry path) | Safe | Uses per-worktree `Mutex<()>` + atomic file writes; thread-agnostic |
| Process registry (`register`/`unregister`) | Safe | Global `Mutex<HashMap>`, no nested locks, no deadlock paths |
| `tail_claude_output` | Safe | 100% synchronous (file I/O + `thread::sleep`); no tokio runtime dependency |
| `ClaudeResponse` across threads | Safe | All fields are `String`/`Vec`/`bool`/`Option` — `Send` is enforced at compile time |
| Cancellation via `cancel_process` | Safe | Registry removal + `SIGKILL` are non-blocking; no cross-lock dependency with spawned thread |

## What is NOT changed

- `execute_claude_detached` function signature and internals — unchanged
- `tail_claude_output` — unchanged
- `ClaudeResponse` type — unchanged
- Process registry — unchanged
- Frontend code — unchanged
- `Cargo.toml` — unchanged (no new dependencies; `tokio::sync::oneshot` is already available)

## Test plan

- [x] `cargo build` — compiles without errors
- [x] `cargo test` — all 93 tests pass
- [ ] Single session: send message, verify streaming and response complete
- [ ] Cancel mid-response: verify clean cancellation
- [ ] 15+ concurrent sessions: verify all stream without hangs
- [ ] Session retry: trigger "session not found" to verify retry path works from OS thread

### DIFF
--- a/src-tauri/src/chat/commands.rs
+++ b/src-tauri/src/chat/commands.rs
@@ -1130,67 +1130,98 @@ pub async fn send_chat_message(
         Some(final_allowed_tools)
     };
 
-    // Execute Claude CLI in detached mode
-    // If resume fails with "session not found", retry without the session ID
-    let mut claude_session_id_for_call = claude_session_id.clone();
-    let (pid, claude_response) = loop {
-        log::trace!("About to call execute_claude_detached...");
+    // Execute Claude CLI in detached mode on a dedicated OS thread.
+    // This prevents tokio thread pool starvation when many sessions run concurrently,
+    // since execute_claude_detached blocks (it tails the output file with thread::sleep).
+    // If resume fails with "session not found", retry without the session ID.
+    let bl_app = app.clone();
+    let bl_session_id = session_id.clone();
+    let bl_worktree_id = worktree_id.clone();
+    let bl_worktree_path = worktree_path.clone();
+    let bl_input_file = input_file.clone();
+    let bl_output_file = output_file.clone();
+    let bl_working_dir = context.worktree_path.clone();
+    let bl_claude_session_id = claude_session_id.clone();
+    let bl_model = model.clone();
+    let bl_execution_mode = execution_mode.clone();
+    let bl_thinking_level = thinking_level.clone();
+    let bl_effort_level = effort_level.clone();
+    let bl_allowed_tools = allowed_tools_for_cli.clone();
+    let bl_parallel_prompt = parallel_execution_prompt.clone();
+    let bl_ai_language = ai_language.clone();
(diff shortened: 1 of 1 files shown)


andrasbacsai:
I will still check your suggestions, thanks for it!

PR #126, opened 2026-02-25
fix(windows): pre-check editor CLI before cmd /c launch

michaelgtallia:
## Summary

- On Windows, `cmd /c code <path>` via `.spawn()` always reports success because `cmd.exe` itself starts fine, even when the wrapped command (`code`, `cursor`) is not in PATH. This caused a silent failure where clicking "Open in VS Code" did nothing and showed no error.
- Added a pre-check using `executable_exists()` (backed by the `which` crate) to verify the editor CLI is in PATH before attempting to launch. If missing, a clear error toast is shown to the user.

## Test plan

- [x] Verified `code` is detected via `which` on Windows
- [x] Confirmed "Open in VS Code" successfully opens VS Code after the fix
- [ ] Test with an editor not installed (e.g. Cursor) to verify the error message appears

### DIFF
--- a/src-tauri/src/projects/commands.rs
+++ b/src-tauri/src/projects/commands.rs
@@ -3053,28 +3053,40 @@ pub async fn open_worktree_in_editor(
     {
         // On Windows, VS Code and Cursor install as .cmd batch wrappers (code.cmd, cursor.cmd).
         // Command::new("code") uses CreateProcessW which can't execute .cmd files directly,
-        // so we wrap them with cmd /c. CREATE_NO_WINDOW prevents cmd.exe console flash.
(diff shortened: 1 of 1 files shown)


michaelgtallia:
Note: this is not a duplicate of [#106](https://github.com/coollabsio/jean/pull/106). That PR fixed the original `NotFound` error by wrapping editor commands with `cmd /c`. However, `cmd /c` via `.spawn()` always reports success because `cmd.exe` itself starts fine, even when the wrapped command is not in `PATH`. This caused a silent failure where the editor never opens and no error is shown. This PR adds a pre-check with `executable_exists()` to catch missing CLIs before attempting the launch.

PR #135, opened 2026-03-01
fix(chat): native clipboard image paste fallback for Linux

karimou5:
## Summary

- **Problem**: On Linux (WebKitGTK), pasting screenshots via Ctrl+V does nothing — WebKitGTK doesn't expose `image/*` items in `clipboardData.items`, so the frontend loop finds no image to process.
- **Fix**: Added a native clipboard fallback using the `arboard` crate. When the web Clipboard API returns no image and no text/html, the app reads the system clipboard directly via Rust, processes the image through the existing pipeline (resize to 1568px max, PNG→JPEG compression), and adds it to the chat.
- Tested on Ubuntu 24.04 (X11) — screenshot paste now works correctly, normal text paste is unaffected.

## Changes

| File | Change |
|------|--------|
| `src-tauri/Cargo.toml` | Add `arboard` dependency with `wayland-data-control` feature |
| `src-tauri/src/chat/commands.rs` | New `read_clipboard_image` Tauri command (~65 lines) |
| `src-tauri/src/lib.rs` | Register the new command |
| `src/components/chat/ChatInput.tsx` | Frontend fallback in `handlePaste` when web API fails |

## How it works

1. User pastes (Ctrl+V) in chat
2. Frontend checks `clipboardData.items` for images (existing behavior)
3. **New**: If no image found AND no text/html in clipboard → calls `read_clipboard_image` (native)
4. Rust reads clipboard via `arboard::Clipboard::get_image()`
5. Converts RGBA → PNG → existing `process_image()` pipeline (resize + compress)
6. Returns `SaveImageResponse` → image appears in chat


andrasbacsai:
Thank you for the PR! 💜

PR #142, opened 2026-03-03
feat(terminal): add iTerm2 support as terminal option

grallc:
## Summary
  - Add iTerm2 as a terminal option alongside Terminal.app, Warp, and Ghostty
  - Uses AppleScript to open a new tab in the current iTerm2 window (or a new window if none exists) and cd into the worktree directory
  - Updates TypeScript types and terminal options list


andrasbacsai:
Thank you for the PR! 💜

PR #148, opened 2026-03-04
feat(notifications): dynamic notification sounds with system & custom support

YusufLisawi:
## Summary

- **Replaced hardcoded sound options** with dynamic system that lists system sounds (macOS alert tones) and custom imported sounds
- **Added backend commands** to list, fetch, import, and delete notification sounds
- **Updated UI** to display sounds grouped by category (default, system, custom) with delete buttons for custom sounds
- **Made playback async** — now resolves sound paths via backend before playing
- **Removed preloading** — sounds load on-demand via asset protocol

## Breaking Changes

- `NotificationSound` type changed from union type (`'none' | 'ding' | 'chime' | 'pop' | 'choochoo'`) to `string` (ID-based: `"none"` | `"system:<name>"` | `"custom:<filename>"`)
- Removed `notificationSoundOptions` static array
- `playNotificationSound()` is now async

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -1491,6 +1491,199 @@ async fn send_native_notification(
     }
 }
 
+// =============================================================================
+// Notification Sounds
+// =============================================================================
+
+/// macOS system alert tones directory
+#[cfg(target_os = "macos")]
+const MACOS_ALERT_TONES_DIR: &str =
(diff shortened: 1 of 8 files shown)


yougotwill:
Was going to make a similar PR myself! Thanks for making it. Hopefully it can be updated and merged 🙏

PR #182, opened 2026-03-10
fix(cli): centralize shell_escape and fix CLI path escaping

### LINKED ISSUE #180: Relogin to the Claude CLI fail on MacOs 15.6.1 (24G90)
I was using the application sucessfuly, but then i installed claude cli to my system using official instalation method.
Then after vacation - (10 days) I had to relogin the claude cli in other applications, and also in Jean it showed again onboarding modal after i opened configuration.

On the onboarding window i select claude cli and on next step i see this issue: 

Image

The path `/Users/lukasbauer/Library/Application` really does not exist, but there are directories `Application Scripts` and `Application Support`
So there is probably some issue with ecaping the space in the directory path.

CoolLamer:
## Summary
- Centralize `shell_escape()` in `platform/process.rs` to handle paths with spaces (e.g., `~/Library/Application Support/...`)
- Replace duplicated `shell_escape` in `detached.rs` with re-export from platform module
- Apply escaping in PTY shell wrapper fallback (`pty.rs`) to fix CLI relogin failures on macOS
- Add warning log when `command_args` is None to surface potential IPC deserialization issues

Fixes #180

### DIFF
--- a/src-tauri/src/chat/detached.rs
+++ b/src-tauri/src/chat/detached.rs
@@ -14,12 +14,9 @@ use std::io::{BufRead, BufReader};
 pub use crate::platform::is_process_alive;
 use crate::platform::silent_command;
 
-/// Escape a string for safe use in a shell command.
+// Re-export shell_escape from platform module
 #[cfg(unix)]
-fn shell_escape(s: &str) -> String {
-    // Use single quotes and escape any single quotes within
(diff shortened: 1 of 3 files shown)


CoolLamer:
Already fixed faster by mantainer of the repo :)

PR #238, opened 2026-03-19
feat(setting): Add settings search

Whiteknight07:
This adds search to Preferences so you can jump directly to panes, sections, keybindings, and magic prompts instead of clicking through each tab manually.

It also smooths out a few details around the new search flow:
- pressing Escape in the search field clears the search without closing the Preferences dialog
- the first search result is highlighted as soon as you start typing so keyboard navigation works immediately

Here is a video demo:

https://github.com/user-attachments/assets/ccb98b4d-b66c-44f9-80d3-18f1a104f9cb

> Note: I am still learning things and it might be possible for this PR to have some issues, I have done my best to ensure no regressions or bugs are caused.


andrasbacsai:
Thank you for the PR! 💜

I added a few fixes here and there.

PR #245, opened 2026-03-20
feat: add per-CLI auto-update on app launch

Whiteknight07:
## Summary
- add per-CLI auto-update preferences for Jean-managed Claude, GitHub CLI, Codex, and OpenCode installs
- run launch-time CLI updates with pinned versions, serialized background installs, and manual-toast fallback on failure
- refactor the CLI version check hook to reduce duplication while keeping the launch-time update flow explicit

## Validation
- bunx eslint src/hooks/useCliVersionCheck.ts
- bun run typecheck

### DIFF
--- a/scripts/build-appimage.sh
+++ b/scripts/build-appimage.sh
@@ -34,7 +34,7 @@ fi
 
 echo "==> Building AppImage via Tauri..."
 cd "$PROJECT_DIR"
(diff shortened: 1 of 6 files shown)


andrasbacsai:
There is already an option for this (enabled by default). Thanks for the PR! 💜

PR #259, opened 2026-03-24
feat(settings): add Remove buttons for optional CLIs

Loa212:
- Add "Remove" button (with trash icon + confirmation dialog) to GitHub CLI, Codex CLI, and OpenCode CLI settings sections
- Rust backend: `uninstall_gh_cli`, `uninstall_codex_cli`, `uninstall_opencode_cli` commands that delete the Jean-managed binary directory
- Commands registered in both native Tauri IPC and WebSocket dispatch
- Removing a CLI resets `default_backend` to Claude if the removed CLI was the active backend
- Button only appears when the CLI is installed; after removal the section shows "Not installed" with an Install button

PS. Claude CLI section has no Remove button (it's mandatory)

### DIFF
--- a/src-tauri/src/codex_cli/commands.rs
+++ b/src-tauri/src/codex_cli/commands.rs
@@ -7,7 +7,7 @@ use std::path::PathBuf;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use tauri::AppHandle;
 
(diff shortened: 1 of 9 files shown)


andrasbacsai:
Jean managed ones could be deleted so this is not needed.

andrasbacsai:
But thanks for the PR! 💜

PR #294, opened 2026-04-12
feat(worktree): allow setting custom branch/worktree name

ngvuhuy:
Normal worktree creation will generate a random name for the worktree and branch. For projects with strict branch naming rules, this means we'd have to rename the branch and the worktree before pushing, which is not the most elegant. Additionally, the random worktree naming is quite confusing when looking around using the terminal.

This PR allows optionally specifying the branch name when creating a new worktree. The worktree name will automatically be derived from this branch name (`feat/worktree-1` -> `feat_worktree-1`)
image


andrasbacsai:
Thank you for the PR! 💜

PR #306, opened 2026-04-15
feat(wsl): add WSL support for Windows CLI tool routing

andrasbacsai:
### Security & Fixes
- WSL-based AI sessions can now be tracked and stopped correctly.

### Improvements
- Windows users can run Jean through a selected WSL distro.
- Jean can install and use Claude, Codex, GitHub CLI, OpenCode, and Cursor inside WSL.
- Windows onboarding now helps choose and validate a WSL setup before selecting AI tools.
- Git, terminals, worktrees, and AI sessions now use WSL paths automatically when WSL mode is enabled.
- Windows settings now include WSL controls for switching mode or changing the selected distro.
- The WSL setup step is now clearly labeled as beta.


Twest2:
I'm super excited for this feature because this is what is currently holding me back from using Jean so I did a full pass on the WSL diff and found a few blocking issues before this is safe to merge:

1. `src-tauri/src/projects/git.rs`: `is_main_worktree()` still shells out with native `git` instead of `wsl_aware_command()`.
   `remove_worktree()` relies on that helper as the safety rail that prevents deleting the primary checkout. In WSL mode we only validate `git` inside the distro, not Git for Windows, so on a machine without native Windows git, or where native git cannot operate on `\\wsl.localhost\...`, `is_main_worktree()` just returns `false` and the guard is bypassed. That makes the main checkout removable in the exact configuration this PR introduces. This path should use WSL-aware git and fail closed if detection fails.

2. `src-tauri/src/projects/git.rs`: `resolve_git_dirs()` has the same native-git leak.
   That helper is used when starting Codex sessions to grant writable access to `.git` / `git-common-dir` roots for worktrees. In WSL mode it can quietly return `None`, which means worktree-backed Codex sessions lose those extra writable roots and can hit permission/sandbox failures again. This should also go through `wsl_aware_command()`.

3. `src-tauri/src/projects/git.rs`: `run_jean_script()` still executes repo setup/teardown scripts through `powershell.exe` on Windows.
   Once WSL mode is enabled, repos and tools live inside the distro, so these scripts are now running under the wrong shell with Windows UNC paths in `JEAN_WORKSPACE_PATH` / `JEAN_ROOT_PATH`. Any repo script that expects `bash`, Linux paths, or WSL-installed tooling will break even though the rest of the repo flow has been moved to WSL. These scripts need the same `wsl.exe -d <distro> --cd <unix_cwd>` handling as the git/CLI paths.

4. `src/components/preferences/panes/GeneralPane.tsx`: the settings flow is less safe than onboarding.
   `WslSetupStep` only persists WSL mode after a distro is selected and `git` has been validated, but the new settings switch immediately saves `{ wsl_enabled: true, wsl_mode_chosen: true }` before any distro exists. That leaves a persisted `enabled + empty distro` state, and every WSL-routed command then becomes `wsl.exe -d "" ...` until the user fixes it manually. The settings path should require a validated distro before enabling WSL, or persist the toggle and distro atomically.

Non-blocking:
- Both WSL preference flows invalidate `claude` / `codex` / `opencode` / `gh` caches, but not `cursor-cli`, even though Cursor detection/auth/model queries are also WSL-sensitive. After toggling WSL, Cursor state can stay stale until a reload.

andrasbacsai:
@coderabbitai review

andrasbacsai:
@coderabbitai review

andrasbacsai:
@coderabbitai review

PR #311, opened 2026-04-16
fix: auto-resize textarea on WebKitGTK (Linux)

karimou5:
## Summary

- Add JS fallback for textarea auto-resize on WebKitGTK (Linux), where CSS `field-sizing: content` is not supported
- New `useAutoResize` hook detects support via `CSS.supports()` — no-op on Chrome/Firefox/Safari, JS resize on WebKitGTK
- Bump `max-height` from `240px` to `50vh` for better multi-line editing
- Hook called at all 9 programmatic value-change sites in `ChatInput.tsx`

## How it works

- If `field-sizing: content` is supported (Chrome 123+, Firefox 131+, Safari 18.4+) → the hook is a complete no-op, CSS handles everything as before
- If not supported (WebKitGTK/Tauri Linux) → the hook resets `height: auto`, reads `scrollHeight`, and caps at `max-height`, toggling `overflowY` as needed
- Listens on `input` + window `resize` events, plus imperative `resizeTextarea()` calls after every programmatic `.value =` assignment

## Test plan

- [x] Type multiple lines → textarea grows line by line
- [x] Delete text → textarea shrinks
- [x] Paste multi-line text → grows up to 50vh then scrolls
- [x] Insert @filename → resizes if text expands
- [x] Enter to send → returns to 1 line
- [x] Switch sessions → resizes for loaded draft
- [x] Verify no regression on Chrome/macOS (field-sizing: content still works natively)

### DIFF
--- a/src/components/chat/ChatInput.tsx
+++ b/src/components/chat/ChatInput.tsx
@@ -1,5 +1,6 @@
 import { memo, useCallback, useEffect, useRef, useState } from 'react'
 import { useIsMobile } from '@/hooks/use-mobile'
+import { useAutoResize } from '@/hooks/use-auto-resize'
 import { invoke } from '@/lib/transport'
 import { generateId } from '@/lib/uuid'
 import { toast } from 'sonner'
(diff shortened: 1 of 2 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #317, opened 2026-04-17
fix(gh-cli): use --active flag for gh auth status check

### LINKED ISSUE #309: gh auth check incorrectly fails when inactive account has stale keyring token
The GitHub CLI authentication check incorrectly reports "not authenticated" even after successfully logging in, if there is more than one account registered and any inactive account has an invalid keyring token.

  ## Steps to Reproduce

  1. Have two GitHub accounts registered with `gh`
  2. The inactive account has a stale/invalid keyring token
  3. Complete `gh auth login` for the active account successfully
  4. Jean still shows the account as not authenticated

  ## Root Cause

  `check_gh_cli_auth` in `src-tauri/src/gh_cli/commands.rs` runs `gh auth status`
  and treats any non-zero exit code as unauthenticated. `gh auth status` exits with
  code 1 if **any** registered account fails — even an inactive one — regardless of
  whether the active account is valid.

  ## Expected Behaviour

  Authentication check should pass if the active account has a valid token.

  ## Actual Behaviour

  Authentication check fails because an inactive account has a stale keyring entry.

  ## Environment

  - Jean version: 0.1.41

xavieroyj:
Fixes auth detection when multiple GitHub accounts are configured by passing --active to only check the currently active account.

Fixes #309

### DIFF
--- a/src-tauri/src/gh_cli/commands.rs
+++ b/src-tauri/src/gh_cli/commands.rs
@@ -665,10 +665,10 @@ pub async fn check_gh_cli_auth(app: AppHandle) -> Result<GhAuthStatus, String> {
     }
 
     // Run gh auth status to check authentication
-    log::trace!("Running auth check: {:?} auth status", binary_path);
+    log::trace!("Running auth check: {:?} auth status --active", binary_path);
 
     let output = silent_command(&binary_path)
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #331, opened 2026-04-25
fix(chat): disable spellcheck on backend model picker search

grallc:
## Summary
- Add `spellCheck={false}` to the search input in `BackendModelPickerContent`.
- Browser autocorrect kept suggesting "got" for "gpt", requiring an extra keypress to dismiss.


andrasbacsai:
Thank you for the PR! 💜

PR #344, opened 2026-04-29
fix(ui): apply cursor-pointer to interactive shadcn primitives

jorgefl8:
## Summary

Noticed that several interactive shadcn primitives don't show a `cursor-pointer` on hover. 

This is a personal opinion that cursor pointer feel more natural and provide a better result for UX.

There were manually  11 cursor pointer to different `<CollapsibleTrigger>` callsites. Fixing the source removes the workaround entirely.

## Root cause

shadcn/ui's defaults are inconsistent on this:
- `Button` has `cursor-pointer` baked in (`button.tsx:8`).
- `Checkbox`, `Switch`, `RadioGroupItem`, `SelectTrigger` only declare `disabled:cursor-not-allowed` — never the enabled-state pointer.
- `Toggle`, `CollapsibleTrigger` have no cursor declaration at all.
- `DropdownMenuItem`, `DropdownMenuCheckboxItem`, `DropdownMenuRadioItem`, `DropdownMenuSubTrigger`, `ContextMenu*` (4 variants), `CommandItem`, and `SelectItem` explicitly set `cursor-default`. This matches macOS native menu conventions but feels off in a web app where users expect "clickable = pointer" (Linear, Vercel, GitHub all override this).

## Fix

In `src/components/ui/`:

- Added `cursor-pointer` to the base className of: `Checkbox`, `Switch`, `RadioGroupItem`, `SelectTrigger`, `Toggle` (CVA base).
- Refactored `CollapsibleTrigger` to accept `className` and apply `cursor-pointer` via `cn()` (it previously didn't merge any class).
- Flipped `cursor-default` → `cursor-pointer` on: `CommandItem`, `SelectItem`, all 4 `DropdownMenu*` items, all 4 `ContextMenu*` items.
- Left `SelectScrollUpButton` / `SelectScrollDownButton` with `cursor-default` on purpose — those are auto-scroll arrows, not items.
- Left `<Label>` untouched: 13 of its 46 callsites are decorative section headings without `htmlFor`, so a global `cursor-pointer` would create false affordances.

In the rest of `src/`:

- Removed the now-redundant manual `cursor-pointer` from 11 `<CollapsibleTrigger>` callsites (`AgentWidget`, `AskUserQuestion`, `PermissionApproval`, `PlanFileDisplay`, `SetupScriptOutput`, `TodoWidget`, plus 5 in `ToolCallInline`).

## Platform impact

- All platforms get the same cursor styling.
- Disabled states keep `disabled:cursor-not-allowed` (later in the cascade, so it wins over `cursor-pointer`). Verified visually on disabled Switch / Checkbox.

## Tests

CSS-only change. I ran the existing test suites for the affected components 29/29 pass. 

## To test

- [ ] Hover a Switch / Checkbox / Radio in Preferences → cursor turns into a pointer.
- [ ] Open a Select dropdown → trigger and items both show pointer.
- [ ] Open a DropdownMenu (e.g. burger menu in chat dock) → items show pointer instead of default arrow.
- [ ] Right-click on something with a ContextMenu → items show pointer.
- [ ] Open a Collapsible (e.g. tool call output in chat) → trigger shows pointer.
- [ ] Hover a Toggle → pointer.
- [ ] Hover a disabled control → cursor stays "not-allowed", not pointer (regression check).
- [ ] Sanity-check that nothing new shows pointer where it shouldn't (e.g. plain text, Label decoratives, Select scroll arrows).


andrasbacsai:
Thank you for the PR! 💜

PR #352, opened 2026-04-30
feat(terminal): user-configurable terminal background color

montagnoli:
## Summary
Adds user-configurable terminal background color so users can match the embedded terminal to their preferred theme/contrast instead of being stuck with the hardcoded default.

## Changes
- New preference for terminal background color (persisted in user settings).
- Settings UI control to pick/reset the color.
- Terminal renderer reads the preference and applies it on mount and on change.
- Falls back to the previous default when unset, so existing users see no behavior change until they opt in.

## Testing
- Open Settings → Terminal → change background color → confirm live update in open terminal tabs.
- Restart app → color persists.
- Reset to default → reverts to original background.


andrasbacsai:
Thank you for the PR! 💜

PR #353, opened 2026-05-01
Format chat durations as seconds or mm:ss

grallc:
## Summary
- keep durations under one minute displayed as seconds only
- display durations of one minute or more in mm:ss format
- add tests for the formatter and MessageItem rendering

## Testing
- bun run test:run src/components/chat/time-utils.test.ts src/components/chat/MessageItem.test.tsx


andrasbacsai:
Thank you for the PR! 💜

PR #355, opened 2026-05-01
fix(terminal): handle split multi-byte UTF-8 codepoints in PTY reader

montagnoli:
## Summary

PTY reader was using `String::from_utf8_lossy` on raw 4096-byte buffer reads. When a multi-byte UTF-8 codepoint was split across two reads (e.g. emoji, CJK, accented chars landing on the buffer boundary), the trailing bytes of the first chunk and the leading bytes of the next chunk were each individually invalid UTF-8 — so `from_utf8_lossy` replaced them with `U+FFFD` (�), corrupting otherwise valid terminal output.

## Fix

Streaming UTF-8 decode in `src-tauri/src/terminal/pty.rs`:

- **Carry incomplete tail**: up to 3 trailing bytes of an unfinished codepoint are stashed and prepended to the next read. The carry is staged at the start of the buffer; the read writes after it — zero-allocation combine.
- **Fast path**: when the combined buffer is fully valid UTF-8, hand bytes straight into a `String` with no scanning/replacement.
- **Slow path**: genuine invalid sequences still produce one `U+FFFD` per bad sequence (matching `from_utf8_lossy` behavior), with a single allocation sized to the input.
- **EOF drain**: any dangling carry at EOF emits one `U+FFFD` per byte, matching `from_utf8_lossy` end-of-stream semantics.

## Repro

Run a command emitting non-ASCII output that produces enough volume to land a codepoint on a 4096-byte boundary (e.g. `cat large-utf8-file`, output containing emoji or CJK at high throughput). Before the fix you'd intermittently see � replacing valid characters; after the fix output is clean.

## Test plan

- [ ] Run a terminal command outputting CJK / emoji at volume; confirm no `U+FFFD` artifacts
- [ ] Run a command emitting genuinely invalid bytes (e.g. `printf '\xff\xfe'`); confirm `U+FFFD` still appears
- [ ] Verify EOF flush: short-lived process emitting incomplete trailing bytes shows `U+FFFD` for those bytes
- [ ] Smoke test ASCII-only output for regressions (fast path)

### DIFF
--- a/src-tauri/src/terminal/pty.rs
+++ b/src-tauri/src/terminal/pty.rs
@@ -161,21 +161,61 @@ pub fn spawn_terminal(
         log::error!("Failed to emit terminal:started event: {e}");
     }
 
-    // Spawn reader thread
+    // Spawn reader thread.
+    //
+    // Streaming UTF-8 decode: a `read()` can split a multi-byte codepoint at
+    // the buffer boundary. `from_utf8_lossy` would emit `U+FFFD` for the split
+    // bytes — corrupting valid output. Instead we carry up to 3 trailing bytes
+    // of an incomplete codepoint into the next read. Genuine invalid sequences
+    // still produce one `U+FFFD` per bad sequence, matching `from_utf8_lossy`.
     let app_clone = app.clone();
     let terminal_id_clone = terminal_id.clone();
     thread::spawn(move || {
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #362, opened 2026-05-06
fix(codex): preserve danger full access for yolo turns

SebastianMihali:
## Summary

Fixes Codex `yolo` turns being accidentally re-sandboxed when Jean sends a per-turn `sandboxPolicy`.

Previously, `thread/start` correctly set yolo sessions to `danger-full-access`, but `turn/start` could override that with `workspaceWrite` whenever extra writable roots were present. This happens for normal Git projects too, because Jean resolves Git metadata directories and passes them as writable roots. As a result, tools that require true unsandboxed execution on macOS, such as Playwright/Chromium headless screenshots, could fail with permission errors even in yolo mode.

This change makes yolo turns explicitly send:

```json
{ "type": "dangerFullAccess" }
```

for `sandboxPolicy`, preserving the intended yolo behavior. Build/plan behavior remains scoped as before.

## Test plan

- [x] `cargo test yolo_turn_uses_danger_full_access_even_with_writable_roots`
- [x] Manually verified that Playwright screenshot flow works in a yolo Codex session after the change


andrasbacsai:
@coderabbitai review

PR #363, opened 2026-05-06
fix(github): avoid false auth prompts for unsupported repos

### LINKED ISSUE #354: GitHub Dashboard shows authentication error for non-GitHub or missing git remotes
## Problem

The GitHub Dashboard can incorrectly show a “GitHub CLI not authenticated” message even when the user is already authenticated with `gh`.

This can happen when one or more configured projects cannot be queried through GitHub CLI because they are not valid GitHub repositories.

Examples include:

- projects hosted on a different provider, such as GitLab
- projects with no git remotes configured
- projects whose remotes do not point to a known GitHub host

## What happens

When opening the GitHub Dashboard, the app queries GitHub data across projects.

For a repository whose remote is not a known GitHub host, `gh` may return an error similar to:

```text
none of the git remotes configured for this repository point to a known GitHub host.
To tell gh about a new GitHub host, please use `gh auth login`
```

For repositories without remotes, `gh` may return errors such as:

```text
no git remotes found
```

Because the first error contains `gh auth login`, the app appears to classify it as an authentication problem and shows the GitHub login prompt.

## Expected behavior

The dashboard should not treat non-GitHub repositories or repositories without git remotes as authentication failures.

Possible behaviors could include:

- skip projects that are not hosted on GitHub
- skip projects with no git remotes configured
- show a per-project warning/error instead of blocking the whole dashboard
- distinguish authentication errors from unsupported/non-GitHub remotes
- only show the login prompt when `gh auth status` actually reports that the user is unauthenticated

## Notes

The user may already be authenticated successfully:

```bash
gh auth status --active
```

So the issue seems to be error classification / repository eligibility detection rather than actual authentication.

SebastianMihali:
## Summary
- Tighten GitHub CLI auth detection so unsupported repositories, missing remotes, and non-GitHub hosts are no longer treated as authentication failures.
- Add shared error classification on the Rust side and reuse it across GitHub issue/PR/security/advisory commands.
- Update the GitHub dashboard to verify real `gh` auth status before showing the login prompt, and surface non-auth command failures separately.
- Add unit tests covering unknown GitHub hosts, missing remotes, and real auth prompts.

## Breaking Changes
- None.

---

Fixes #354


andrasbacsai:
@coderabbitai review

andrasbacsai:
@coderabbitai review

andrasbacsai:
Thank you for the PR! 💜

PR #371, opened 2026-05-15
perf(macos): reduce GPU load with reduced blurring/animations, shared tick store and optional vibrancy

petrbela:
Reduced GPU load on MBP M1 Pro from consistent ~75% to around 10-30% depending on load. If needed I could probably run some profiling numbers.

Also, I'm not sure if all the changes are necessary for the fix (the first two were the major reduce; the TickStore might not be needed) but I didn't notice any degradation in the UI so I'm submitting it as is. I've been using this version for a few weeks now so it should be reasonably stable...

## AI Summary

- **Eliminate `backdrop-blur` from UI elements**: Removes `backdrop-blur-sm/md` from dialogs, floating buttons, title bar, floating dock, and overlay components — the main source of persistent GPU compositor overhead
- **Opacity-based animations**: Converts `icon-glow` CSS animation from `filter: drop-shadow` (GPU-expensive) to `opacity` transitions; adds `prefers-reduced-motion` support and pauses all animations when the window loses focus via a `window-blurred` class
- **Shared tick store**: Introduces `src/contexts/TickStore.ts` — a single `setInterval`-backed store consumed via `useSyncExternalStore`, replacing per-component `useState` + `setInterval` patterns in `useElapsedTime` and `useWakeupRemaining`; the interval is automatically paused when the document is hidden or window is blurred
- **Optional macOS window vibrancy**: Adds a `window_vibrancy` preference (default: `false`) with a toggle in Appearance settings (macOS only); uses `tauri::window::Effect::Sidebar` + `objc2-app-kit` to set window transparent/opaque at runtime; on startup, explicitly marks the window opaque to prevent unnecessary compositor blending for users who haven't opted in
- **New Tauri command** `set_window_vibrancy`: Exposed via `dispatch.rs` and `invoke_handler`, applies or clears vibrancy effects live without restart


petrbela:
@andrasbacsai Maybe a few things to double check, I've noticed two issues recently which I'm not sure are intended:
1. display of how long the current agent has been running for doesn't update in real time when the window is not focused (expected given the change here)
2. it happened to me yesterday that a branch I was working on got a push from another user, my Jean client never pulled it, and then "Commit and Push" instead of merging the change, it instead force pushed, skipping the other user's commit (this one I'm not sure if it's preexisting in main?)

Perhaps I should revert changes to the TickStore and only keep the animation optimizations?

Image 21 05 2026 at 10 56

andrasbacsai:
> @andrasbacsai Maybe a few things to double check, I've noticed two issues recently which I'm not sure are intended:
> 
> 1. display of how long the current agent has been running for doesn't update in real time when the window is not focused (expected given the change here)
> 2. it happened to me yesterday that a branch I was working on got a push from another user, my Jean client never pulled it, and then "Commit and Push" instead of merging the change, it instead force pushed, skipping the other user's commit (this one I'm not sure if it's preexisting in main?)
> 
> Perhaps I should revert changes to the TickStore and only keep the animation optimizations?
> 
> Image 21 05 2026 at 10 56

Yes please, revert it. Only optimisations should be here.

andrasbacsai:
@coderabbitai review

andrasbacsai:
Thank you for the PR! 💜

PR #372, opened 2026-05-15
fix: gate RunEvent::Reopen on macos

jlucaso1:
`RunEvent::Reopen` is a macOS-only variant in Tauri 2, so the match arm fails to compile on Linux and Windows; this caused the `v0.1.48` release workflow to ship only macOS bundles.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -3853,6 +3853,7 @@ pub fn run() {
                 }
                 chat::codex_server::shutdown_server();
             }
+            #[cfg(target_os = "macos")]
             tauri::RunEvent::Reopen { .. } => {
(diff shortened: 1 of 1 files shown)


jlucaso1:
Justo to context, v0.1.48 has only builds assets for macos, https://github.com/coollabsio/jean/releases/tag/v0.1.48. Here is the proof in the CI: https://github.com/coollabsio/jean/actions/runs/25797193065.
Has introduced in that commit: https://github.com/coollabsio/jean/commit/65161c82214b022b86baef90b0d826c481c89340

CC: @andrasbacsai

andrasbacsai:
Thank you. It has been fixed in the latest version that is releasing soon.

PR #376, opened 2026-05-15
[codex] fix OpenCode cancel cleanup

### LINKED ISSUE #375: OpenCode cancel can leave session with active request
## Problem

When using the OpenCode backend, cancelling the first prompt in a new session before OpenCode responds can leave the Jean session stuck. Sending another prompt in the same session then fails with:

```text
Session already has an active request
```

## Reproduction

1. Create a new Jean chat session using OpenCode as the backend.
2. Send any prompt.
3. Cancel from the UI before OpenCode returns a response.
4. Send a second prompt in the same session.

## Expected behavior

The cancelled OpenCode request should be cleaned up and a new prompt should start normally.

## Actual behavior

Jean rejects the second prompt because the session is still considered actively managed.

## Root cause

The OpenCode execution path registers a cancel flag in Jean's chat registry and sends the prompt via a long-running blocking HTTP `POST /session/{id}/message` request. Cancelling sets the flag and emits `chat:cancelled`, so the UI clears immediately, but backend registry cleanup only runs after the blocking POST returns. While it is still blocked, `CANCEL_FLAGS` remains registered and the next send hits Jean's duplicate-send guard.

This is OpenCode-specific because Claude/Codex use process or turn interruption paths that unblock cleanup more directly.

## Proposed fix

Make the OpenCode `/message` request locally cancellable, keep sending OpenCode's server-side `POST /session/{id}/abort`, and ensure Jean returns a cancelled response promptly so registry cleanup clears the active request state.

horacioh:
## Summary

- make the OpenCode `/session/{id}/message` POST locally cancellable from Jean
- keep sending OpenCode's server-side `/abort` request, now via a centralized helper
- add regression coverage for OpenCode cancellation cleanup and cancel polling

## Root cause

OpenCode sends prompts through a long-running HTTP POST. When the UI cancelled a run, Jean emitted `chat:cancelled` and set the OpenCode cancel flag, but registry cleanup could not run until the blocking POST returned. The session stayed in `CANCEL_FLAGS`, so the next prompt hit Jean's duplicate-send guard and failed with `Session already has an active request`.

## Validation

- `bun run build`
- `cargo test --manifest-path src-tauri/Cargo.toml await_opencode_http_or_cancel_returns_none_when_cancelled`
- `cargo test --manifest-path src-tauri/Cargo.toml cleanup_session_registrations_clears_opencode_active_flag`
- `git diff --check && cargo fmt --manifest-path src-tauri/Cargo.toml --check`

Fixes #375

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -50,7 +50,7 @@ portable-pty = "0.8"  # For terminal/PTY support
 which = "7"           # For cross-platform executable detection
 axum = { version = "0.8", features = ["ws"] }  # HTTP server + WebSocket
 tower-http = { version = "0.6", features = ["cors", "fs", "compression-br", "compression-gzip"] }  # CORS + static files + response compression
-tokio = { version = "1", features = ["sync", "macros", "net", "io-util", "fs", "rt-multi-thread"] }  # Channel for WS broadcast
+tokio = { version = "1", features = ["sync", "macros", "net", "io-util", "fs", "rt-multi-thread", "time"] }  # Channel for WS broadcast
 futures-util = "0.3"  # Stream utilities for WebSocket split
 image = { version = "0.25", default-features = false, features = ["png", "jpeg", "gif", "webp"] }  # Image resize/compression on paste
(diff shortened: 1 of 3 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #382, opened 2026-05-21
feat(projects): add manual worktree ordering

horacioh:
https://github.com/user-attachments/assets/22a34e4b-2041-450f-b0a5-b9a648fe1f55

https://github.com/user-attachments/assets/e57b0b4b-7c20-4c20-84cd-c30b1e12d0cf

## Summary
- Adds a Manual worktree sort mode to the project canvas selector and shared sort-mode types.
- Enables drag-and-drop reordering in the worktree list, switches the project to manual sorting after a reorder, and persists the new order through the existing reorder mutation.
- Keeps base sessions pinned at order 0 while assigning contiguous order values only to reorderable worktrees.
- Adds tests for manual sorting behavior and store support for the new mode.


horacioh:
@andrasbacsai I wanted to move from dnd-kit to pragmatic-drag-and-drop on this, because I believe is a better/lightweight library. I decided to leave this migration separate from this.

Can we consider it? Do you have any preferences for dnd-kit over something else?

andrasbacsai:
> @andrasbacsai I wanted to move from dnd-kit to pragmatic-drag-and-drop on this, because I believe is a better/lightweight library. I decided to leave this migration separate from this.
> 
> 
> 
> Can we consider it? Do you have any preferences for dnd-kit over something else?

I don't have preference. If that is more lightweight then use that. 🙂

andrasbacsai:
Thank you for the PR btw! 💜

horacioh:
changed the PR to also change from dnd-kit to pragmatic-drag-and-drop :)

andrasbacsai:
I added a few fixes here and there, so it is smoother.

PR #390, opened 2026-06-04
fix: hide seen failed workflow runs

jibin7jose:


andrasbacsai:
Thanks for the PR, but without any details in the description, I won't merge this.

PR #392, opened 2026-06-05
fix: active chat provider switch

jibin7jose:


andrasbacsai:
Thanks for the PR, but without any details in the description, I won't merge this.

PR #411, opened 2026-06-11
fix(terminal): accented characters duplicated in embedded terminal on Linux

fsioni:
## Problem

On Linux, typing accented characters (`é`, `ç`, `à`…) in the embedded terminal produces a growing stream of duplicated characters: typing `ééé` yields `é`, `é`+`é`, `é`+`éé`… The corruption happens in a bare shell too (no Claude Code involved) and only in the embedded terminal — native terminals like Ghostty are fine.

## Root cause

WebKitGTK (Tauri's Linux webview) with an ibus input method commits composed characters by firing `compositionend` **without a preceding `compositionstart`**. The keystroke arrives as:

```
keydown  key="Unidentified" keyCode=229
input    inputType=insertFromComposition data="é"
compositionend data="é"          ← no compositionstart, ever
```

xterm.js's `CompositionHelper` assumes balanced composition events. With no `compositionstart`, its internal composition start position stays stuck at 0, so every orphan `compositionend` re-sends the hidden textarea's **accumulated** content (the textarea is only cleared on Enter/Ctrl+C). The character itself is already delivered once, correctly, through xterm's `keydown(229)` textarea-diff path (`_handleAnyTextareaChanges`), which is why each keystroke yields one correct char plus the whole history again.

Root cause confirmed by instrumenting a standalone WebKitGTK window running the same xterm.js v6 build and tracing real keyboard events (trace excerpt above).

## Fix

`attachOrphanCompositionEndGuard()` — a capture-phase listener pair on the terminal host element (an ancestor of xterm's hidden textarea, so it runs before xterm's own listeners) that swallows `compositionend` events with no matching `compositionstart`:

- Balanced sequences (real IME preedit — CJK, etc.) pass through untouched.
- The guard is therefore a no-op on unaffected platforms — no OS gating needed.
- Wired in `attachToContainer()` after `terminal.open()` for the xterm renderer only, cleaned up on dispose (same lifecycle as `touchScrollCleanup`).

The second commit is a prettier-only reformat of `AutoFixPane.test.tsx` that was blocking `format:check` on this branch.

## Testing

- 5 regression tests in `terminal-composition-guard.test.ts` (orphan swallowed, balanced pairs untouched, orphan-after-pair, repeated-orphan é→éé→ééé scenario, cleanup).
- Verified end-to-end with real keystrokes in a WebKitGTK harness: before the guard each keystroke re-sent accumulated input; with the guard, exactly one character per keystroke.
- `typecheck`, `lint`, `format:check`, full TS suite (826 tests / 137 files) green.

Upstream xterm.js issue candidate: `CompositionHelper` does not tolerate orphan `compositionend` events (WebKitGTK+ibus emits them for every composed single char).


fsioni:
Addressed both Copilot review comments in 7cba0833:

- **Per-element composition tracking** — the guard now records the `event.target` of `compositionstart` and only treats a `compositionend` as balanced when it targets that same element (instead of a single boolean for the whole root). A `compositionstart` on one descendant can no longer let an orphan `compositionend` from another element through.
- **Re-run-safe attach** — `attachToContainer` now calls the prior `compositionGuardCleanup?.()` and resets it to `null` before (re)attaching, so a reattach can't leak listeners on a stale host or register duplicate guards.
- Added a regression test for the target-mismatch orphan path (6 guard tests total).

fsioni:
## Residual intermittent duplication — root cause + fix (82adde80)

The original guard fixed the systematic `é → éé → ééé` re-send, but an intermittent duplication remained (an extra `é` showing up occasionally, load-dependent, unreproducible on demand). Root cause, proven on the vendored xterm v6 source:

With orphan `compositionend`s swallowed, a composed char's **only** delivery path is `CompositionHelper._handleAnyTextareaChanges` — snapshot `textarea.value` on `keydown(229)`, send the diff in a `setTimeout(0)` (`_inputEvent` ignores `insertFromComposition`; it only handles `insertText`). ibus is asynchronous, so the `keydown/input/compositionend` triplets of consecutive keystrokes can arrive in one burst (DBus round-trip latency, main-thread jank) before any timer runs:

```
kd1 snap "x" → in1 "xé" → kd2 snap "xé" → in2 "xéé"
→ timer1 diff = "éé", timer2 diff = "é"   ⇒ "ééé" for two keystrokes
```

The accumulated textarea (only cleared on Enter/^C/blur) is what the stale snapshots alias against. This also explains why the slow synthetic-keystroke harness validated the original fix: no burst, no race.

### Fix

The guard now takes over delivery at the `input` event: an `insertFromComposition` **outside any open composition** (the WebKitGTK+ibus orphan signature) has its data stripped from the textarea and delivered exactly once via `terminal.input(data, true)`. xterm's racy diff never sees composed chars anymore and the textarea no longer accumulates them. Balanced sequences (real IME preedit — CJK etc.) still pass through untouched, so the guard remains a no-op on unaffected platforms.

Side benefit: also fixes the symmetric race where the diff timer fired *before* the async ibus commit landed, which made the char show up one keystroke late (likely the "à doesn't seem fixed" sightings).

7 new regression tests (burst race, timer-before-commit, balanced IME untouched, `insertText` untouched, suffix-mismatch fallback, no-callback legacy behavior, cleanup). `typecheck`, `lint`, `format:check`, full TS suite (137 files / 834 tests) green.

Upstream note: the `_handleAnyTextareaChanges` snapshot/diff race is reproducible in stock xterm.js whenever consecutive 229-routed keystrokes outpace the 0 ms timer — worth mentioning in the upstream issue alongside the orphan `compositionend` tolerance.

fsioni:
## Field testing showed 82adde80 was still insufficient — full root cause now trace-proven + fix (bab114a3)

After shipping the burst-race fix, real-world testing on Linux (WebKitGTK + ibus, AZERTY) still produced corruption: bursts of stale spaces/letters appearing right after an accent (`corrigé     .`), and accents occasionally being erased right after they were typed. I instrumented a local build with a DOM/PTY event tracer and captured the failures live. Three facts came out of the traces:

**1. WebKitGTK's input method inserts text into xterm's hidden textarea even for keys xterm already handled and `preventDefault()`ed.** Every typed space and shifted letter accumulates there (xterm only clears the value on Enter/^C/blur). Bonus quirk: a typed space is sometimes inserted as **NBSP** while `event.data` reports `" "`.

**2. The flush vector is `CompositionHelper._handleAnyTextareaChanges`** (scheduled on every `keydown` 229). Its three branches were all caught in the act:
- *grew* → re-sends the diff (the original `éé` duplication — already fixed);
- *shrank* → sends `DEL`, **erasing the accent that was just delivered** (WebKit's stale composition range can make a commit REPLACE the textarea's last char instead of appending);
- *same length but different content* → **sends the ENTIRE textarea value to the PTY**. This is the burst: when a composed char commits right after a trailing NBSP, WebKit's editing layer normalizes that NBSP to a plain space — same length, different content — and the whole accumulated junk goes to the shell. Trace: burst payload length matched `textarea.value.length` exactly, ~45-50ms after each commit (the `setTimeout(0)`).

**3. The previous suffix-strip was structurally unable to cope**: the UA mutation isn't always an append (replace + NBSP rewrite), and the inserted char doesn't always equal `event.data` (NBSP vs space).

### Fix
The guard now snapshots `textarea.value` on `beforeinput` and restores it **verbatim** on `input`:
- orphan `insertFromComposition` commits: restore + deliver exactly once via the callback (as before);
- `insertText` echoes of keys xterm already delivered directly (last keydown ≠ 229, no open composition): restore only — this drains the accumulation at the source.

The diff timer then always sees `oldValue === newValue`, so none of its branches can ever fire. Keys typed while an IME is active (`keydown` 229 + `insertText`) are deliberately left on xterm's grow-diff path, which is the only thing that delivers them; balanced compositions (CJK preedit, dead-key `ê` on this platform) remain untouched.

17 unit tests including regressions for the NBSP-normalization burst, the replace→DEL erasure, and the space-accumulation drain. Full TS suite green (137 files / 838 tests).

fsioni:
Field validation: a full day of normal use on the affected setup (Linux, WebKitGTK + ibus, AZERTY, Claude Code TUI in the embedded terminal) with zero corruption — no duplicated accents, no junk-space bursts, no erased characters. Trace instrumentation confirmed the mechanism: every IM echo restored, no diff-timer branch ever fired, hidden textarea stays drained (max 2 inert chars vs 29 before).

fsioni:
Update after 4 days of daily use: the fix has been rock solid — accented characters (é, ç, à…) now type exactly once in the embedded terminal on Linux, with no duplication or accumulation. No regressions observed on real IME input (composed sequences still work fine). Everything looks perfect now. 🎉

andrasbacsai:
Thank you for the PR! 💜

PR #439, opened 2026-06-25
fix(projects): merge PRs with a method the repository allows

fsioni:
## Problem

`merge_github_pr` hard-codes `gh pr merge --merge`. On repositories that
disable merge commits (a common GitHub setting — squash-only or rebase-only
repos), `gh pr merge --merge` exits with code 1
(`Merge commits are not allowed on this repository`).

The mergeability pre-check only rejects `CONFLICTING` / `BLOCKED`, so a clean,
mergeable PR passes the check and then fails at merge time. This affects the
chat "Merge PR" action and the Magic modal merge.

## Fix

Before merging, resolve a method the repository actually allows via:

```
gh repo view --json squashMergeAllowed,mergeCommitAllowed,rebaseMergeAllowed
```

and pass the matching flag (`--squash` / `--merge` / `--rebase`). Falls back to
`--squash` when the settings can't be read. No new commands or dependencies.

## Design note

The implementation prefers **squash > merge > rebase**. This changes the default
only for repos that allow *both* merge and squash commits (they now squash). If
preserving the previous merge-first default matters, the order is a one-line
change — happy to switch to merge > squash > rebase.

## Testing

- `cargo check`, `cargo clippy --lib` (0 warnings), `cargo fmt --check`
- Existing lib tests pass


andrasbacsai:
I added as normal merge as fallback, otherwise it is good.

Thank you for the PR! 💜

PR #459, opened 2026-06-29
fix(rtk): install CLI from verified release asset

### LINKED ISSUE #429: [bug] rtk install button within Settings / Opinionated doesn't work for Windows
On Windows, in Jean desktop app.
I go to Settings, Opinionated, click on Install RTK. Have the message: 
"Fail to install RTK: Fail to run installer: program not found."

After some digging, it seems that `install_rtk()` in https://github.com/coollabsio/jean/blob/c068a5ed3e2625f0d346e414380059592bf2aa9f/src-tauri/src/opinionated/commands.rs expect sh.exe in PATH and fails otherwise.

I have sh.exe because I have installed git-bash (https://git-scm.com/install/windows). But the issue is that the git installer doesn't recommend to check "Use Git and optional Unix tools from the command prompt" and has a recommended default check "Git from the command line" only.

And that's normal because sh.exe won't work as expected on Windows and even in the PATH, the Install RTK in Jean will fail, with this message: "FAILED to install RTK: RTK installation failed:"

The only way to get out of that is to install rtk outside of Jean, manually, going to catch the latest release buried into https://github.com/rtk-ai/rtk/releases, which is not convenient for beginners.

Refers:
- #201
- #163

andrasbacsai:
## Summary
- Replaces the RTK install flow with direct downloads from the latest GitHub release asset for the current platform.
- Verifies downloaded archives against published checksums before installing the managed RTK binary.
- Adds Windows-friendly PATH handling and install verification so Settings > Opinionated works without requiring `sh.exe` or Git Bash Unix tools.

## Tests
- Added coverage for platform asset selection, checksum parsing, zip extraction, and PATH update behavior.

fixes #429

---

Fixes #429

### DIFF
--- a/src-tauri/src/opinionated/commands.rs
+++ b/src-tauri/src/opinionated/commands.rs
@@ -1,7 +1,10 @@
 use crate::platform::silent_command;
 use serde::Serialize;
+use sha2::{Digest, Sha256};
+use std::ffi::{OsStr, OsString};
+use std::io::Read;
 use std::path::{Path, PathBuf};
-use tauri::AppHandle;
+use tauri::{AppHandle, Manager};
 
 #[derive(Debug, Serialize)]
 pub struct PluginStatus {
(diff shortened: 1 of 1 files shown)


andrasbacsai:
@coderabbitai review

Issue #460, opened 2026-06-29
AskUserQuestion silently degrades to plain-text on Jean-managed Claude CLI ≥ 2.1.187


azeitler:
## Summary

With `claude_cli_source = jean` (default), the Claude CLI no longer offers the `AskUserQuestion` tool to the model. Plan-mode clarifications and skills that depend on it silently fall back to plain-text "reply with a number" prompts — no error. Jean's renderer (`src/components/chat/AskUserQuestion.tsx`) already supports the tool, so this is a tool-availability regression, not a rendering bug.

## Root cause

Intentional upstream change. Bisecting bundled CLI init tool lists: `AskUserQuestion` is present **through 2.1.186**, gated out from **2.1.187** (clean consecutive-version boundary), still absent in 2.1.191. Upstream converted it to a **host-dialog** that the Agent SDK exposes only when the embedder declares `supportedDialogKinds` **and** provides an `onUserDialog` handler ([go SDK validates both are required](https://github.com/Roasbeef/claude-agent-sdk-go)). Jean wires neither.

Refs: [py-sdk#327](https://github.com/anthropics/claude-agent-sdk-python/issues/327) · [ts-sdk#109](https://github.com/anthropics/claude-agent-sdk-typescript/issues/109) · [working impl: ben-vargas/ai-sdk-provider-claude-code](https://github.com/ben-vargas/ai-sdk-provider-claude-code)

## Why it surfaced now

`auto_update_ai_backends` (default `true`, added 2026-05-06) silently background-updates the managed CLI on startup, so installs roll past 2.1.186 unprompted. The old `versions.truncate(5)` cap (`src-tauri/src/claude_cli/commands.rs:293`) then drops 2.1.186 from the pin list — so the prior stopgap (pin / `path` to ≤ 2.1.186) no longer works.

## Proposed fix

Jean spawns the CLI in `--print --output-format stream-json --input-format stream-json` (`src/chat/claude.rs:294-298`), so the dialog must be handled over the **stream-json control channel**, not an SDK options object:

1. Advertise `supportedDialogKinds` on the control-protocol init message.
2. On inbound `request_user_dialog` / `permission_ask_user_question`: render via the existing `AskUserQuestion.tsx`, return the answer over the control channel.
3. Map dialog-kind variants (`ask_user_question` / `askUserQuestion` / `AskUserQuestion`) defensively.

Stopgap: raise/remove the `versions.truncate(5)` cap so ≤ 2.1.186 can be pinned again.

[...]

PR #461, opened 2026-06-29
fix(claude-cli): keep more version history, default auto-update off (#460)

azeitler:
## What

Partial mitigation for #460 (`AskUserQuestion` silently degrades to plain-text on Jean-managed Claude CLI ≥ 2.1.187).

Two small changes:

- **Raise managed Claude CLI version cap 5 → 12** (`src-tauri/src/claude_cli/commands.rs`). The 5-version cap dropped 2.1.186 — the last release that exposes `AskUserQuestion` — off the selectable list, so users couldn't pin it. 12 keeps enough history to pin a slightly older release.
- **Default `auto_update_ai_backends` to `false`** (`src-tauri/src/lib.rs`, `src/types/preferences.ts`). Auto-update silently background-installs the latest stable on startup, rolling the managed CLI past a working/pinned version unprompted. Defaulting off surfaces updates via toast instead; users opt in via Settings.

## What this does NOT fix

The real fix — wiring the stream-json control-channel dialog broker (`supportedDialogKinds` + `onUserDialog`) so `AskUserQuestion` works on ≥ 2.1.187 — is still open in #460. This PR only stops the silent roll-forward and restores the pin-an-older-version stopgap.

## Notes

- `auto_update_ai_backends` is a single global preference covering all backends (Claude, Codex, gh, opencode, pi, coderabbit), so this flips the default for all of them. Existing users with the pref already persisted are unaffected; only new/default installs change.

Refs #460

### DIFF
--- a/src-tauri/src/claude_cli/commands.rs
+++ b/src-tauri/src/claude_cli/commands.rs
@@ -293,8 +293,10 @@ pub async fn get_available_cli_versions() -> Result<Vec<ReleaseInfo>, String> {
         b_parts.cmp(&a_parts)
     });
 
-    // Take only the 5 most recent versions
-    versions.truncate(5);
+    // Take only the 12 most recent versions. Keeps enough history that users
+    // can pin a slightly older release (e.g. one that still exposes the
+    // AskUserQuestion tool) instead of being forced onto the very latest.
+    versions.truncate(12);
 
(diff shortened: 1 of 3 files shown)


andrasbacsai:
I do not like to make auto-update off by default, that could cause security and other issues. I think it is better that you turn it off and you should able set the correct version.

PR #464, opened 2026-06-30
feat(commands): add active session rename command

### LINKED ISSUE #359: Renaming active panes during activity terminates, preventing rename.
There is an inconsistency in the pane renaming functionality. Currently, the system allows renaming only when a pane is idle. If a pane is actively running a task, the rename action terminates. Additionally, attempting to rename an active pane from a separate inactive pane also terminates.

Users should be able to rename any pane regardless of its current activity state.

andrasbacsai:
## Summary
- Add a command palette action for renaming the current session tab.
- Allow rename requests to target the selected worktree/session even when the chat pane is not the active pane.
- Keep the rename input interactive without triggering tab selection, close, drag, or cancel behavior.

closes #359

---

Fixes #359


andrasbacsai:
@coderabbitai review

PR #494, opened 2026-07-19
Fix web terminal fast-typing input ordering

toanalien:
## Issue

In Jean web access, the embedded xterm.js terminal could display typed characters in the wrong order when the user typed quickly. For example, typing `hello world` could sometimes appear as `hewollo ld`.

The visible symptom looked like terminal lag or xterm.js processing keystrokes incorrectly, but the terminal UI was not the root cause. xterm.js emitted input chunks in order through `onData`; the ordering problem happened after Jean forwarded those chunks to the backend PTY.

## Root cause

In web mode, terminal input is sent over the WebSocket transport as `terminal_write` invocations.

Before this fix, every WebSocket `invoke` was dispatched through its own async task. That worked for ordinary commands, but it was unsafe for terminal input: fast typing can produce multiple small input chunks, and each chunk could race independently to acquire the PTY writer lock.

Example input flow:

```text
xterm.js onData emits: "he" -> "llo" -> " wo" -> "rld"
WebSocket receives:      "he" -> "llo" -> " wo" -> "rld"
Spawned tasks write:     "he" -> " wo" -> "llo" -> "rld"
Terminal displays:       "hewollo rld" / similar corruption
```

The WebSocket preserved receive order, but Jean lost that ordering by spawning the writes as independent tasks.

## Solution

`terminal_write` now runs inline inside the WebSocket connection loop instead of being spawned as a separate task.

This keeps PTY writes in the same order as the WebSocket messages received from the browser. Other commands still use the existing spawned-dispatch path so long-running or unrelated commands do not block event delivery.

Implementation details:

- Added `command_must_run_in_ws_order("terminal_write")`.
- Routed ordered commands through `dispatch_invoke_response(...).await` directly in the WebSocket loop.
- Kept legacy invoke handling consistent with the typed WebSocket message path.
- Added a regression test proving `terminal_write` is treated as an ordered command while commands like `terminal_resize` and `create_session` are not.

## Scope

This fix intentionally targets the common single-browser-connection case that caused fast typing to be reordered in Jean web access.

It does not introduce a per-`terminal_id` backend input queue for multi-tab or multi-client shared terminal editing. That would be a larger architectural change and should be handled separately if Jean decides to support stronger
multi-client terminal input semantics.

## Test plan

- [x] `cargo test --manifest-path jean-core/Cargo.toml http_server::websocket::tests`
- [x] `cargo test --manifest-path jean-core/Cargo.toml http_server::`
- [x] `cargo test --manifest-path jean-core/Cargo.toml terminal::`
- [x] `git diff --check`

Manual smoke test:

1. Open Jean web access in a browser.
2. Focus an xterm.js terminal.
3. Type `hello world` quickly several times.
4. Confirm the terminal echoes characters in the same order typed.

CleanShot 2026-07-19 at 12 13 42@2x

CleanShot 2026-07-19 at 12 14 03@2x

CleanShot 2026-07-19 at 12 14 23@2x

### DIFF
--- a/jean-core/src/http_server/websocket.rs
+++ b/jean-core/src/http_server/websocket.rs
@@ -33,6 +33,10 @@ fn command_should_run_on_blocking_pool(command: &str) -> bool {
     )
 }
 
+fn command_must_run_in_ws_order(command: &str) -> bool {
+    matches!(command, "terminal_write")
+}
+
 async fn dispatch_invoke_response(
     app: AppHandle,
     id: String,
@@ -88,7 +92,7 @@ fn spawn_dispatch_response(
 
 #[cfg(test)]
 mod tests {
-    use super::command_should_run_on_blocking_pool;
+    use super::{command_must_run_in_ws_order, command_should_run_on_blocking_pool};
 
     #[test]
     fn commit_generation_runs_on_blocking_pool() {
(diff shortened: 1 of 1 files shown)


toanalien:
Hello @andrasbacsai, do you have time to review this PR? Thank you!

andrasbacsai:
Thank you for the PR! 💜

PR #497, opened 2026-07-20
fix(terminal): preserve unicode input in web access

toanalien:
## Summary

- Fix Web Access terminal Unicode corruption when typing Vietnamese in xterm.js/zsh sessions.
- Add a Unix PTY locale fallback that sets `LC_CTYPE` to a UTF-8 locale only when the parent environment is not already UTF-8 and has not explicitly set `LC_ALL` or `LC_CTYPE`.
- Add regression coverage for the terminal locale fallback behavior.
- Apply required Rustfmt/Clippy cleanups so Rust quality gates pass.

## Root cause

Vietnamese input using decomposed combining marks could be corrupted by zsh when the PTY process started with a non-UTF-8 locale. The visible symptom was replacement/control-byte artifacts such as `<0080>` in Web Access terminal input. The terminal output decoder already preserves split UTF-8 output, so the fix belongs at PTY process locale setup.

## Verification

- `cargo test --manifest-path jean-core/Cargo.toml terminal_`
- `cargo test --manifest-path jean-core/Cargo.toml`
- `bun run rust:fmt:check`
- `bun run rust:clippy`
- `bun run rust:test`

## Manual test plan

1. Start Jean Web Access.
2. Open an xterm terminal.
3. Run `locale` and confirm `LC_CTYPE` is UTF-8.
4. Type `xin chào mẹ` using the Vietnamese input method.
5. Press Enter and confirm no `�`, `<0080>`, or other control-byte artifacts appear.

### DIFF
--- a/jean-core/src/chat/codex.rs
+++ b/jean-core/src/chat/codex.rs
@@ -3686,14 +3686,13 @@ pub(crate) fn codex_run_log_has_visible_assistant_artifacts(
             .unwrap_or("")
         {
             "turn.plan_updated" | "item.plan.delta" if is_plan_mode => return true,
-            "turn.completed" => {
+            "turn.completed"
                 if msg
                     .get("output")
                     .and_then(extract_text_from_turn_output)
-                    .is_some_and(|text| !text.trim().is_empty())
-                {
-                    return true;
-                }
+                    .is_some_and(|text| !text.trim().is_empty()) =>
+            {
+                return true;
             }
             "item.started" | "item.completed" => {
                 let item = msg.get("item").unwrap_or(&serde_json::Value::Null);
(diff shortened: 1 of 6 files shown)


andrasbacsai:
Thanks for this fix — the PTY `LC_CTYPE` UTF-8 fallback is the right approach.

This branch (from the fork) had fallen far behind `main` and also included unrelated clippy/rustfmt-only cleanups. I've rebased the locale fix onto current `main`, dropped the incidental edits, expanded the locale unit tests, and opened a refreshed PR:

**#606** — https://github.com/coollabsio/jean/pull/606

Please close this PR if #606 looks good, or re-point this branch if you prefer to keep ownership on the fork.

toanalien:
@andrasbacsai  Thank you for reviewing this PR, your feedback was really helpful. I see #606 looks good, so I’ll go ahead and close this one.

PR #500, opened 2026-07-20
feat: add GitLab as an alternative git provider

### LINKED ISSUE #150: [Feature] Add GitLab as an alternative to GitHub
I’d like to request GitLab support as a second provider option, so GitLab users can use the same Jean workflow without being locked to GitHub.

GitLab provides its own [CLI](https://docs.gitlab.com/cli/)

mgvdev:
Implements #150 — GitLab support as an alternative git provider, alongside the existing GitHub integration.

## Approach

Rather than bolting GitLab on as a parallel feature, this introduces a **git provider abstraction**. Each project resolves to a provider, and the existing commands dispatch on it — so the frontend, MCP tools and HTTP/WS dispatch keep using the same command names and response shapes.

Provider resolution precedence:
1. A `provider` block in the repo's `jean.json` (explicit, committed)
2. Auto-detection from the `origin` remote (`github.com` / `gitlab.com` / self-hosted hosts containing `gitlab`)
3. Fallback to GitHub (preserves existing behaviour)

```jsonc
// jean.json
"provider": { "git": "gitlab", "host": "gitlab.example.com" }
```

No secrets live in `jean.json` — only the (non-secret) provider selection and host. Each CLI manages its own auth, exactly like `gh` does today.

## Main changes

**Provider model**
- New `GitProvider` enum with a tolerant deserializer (unknown values fall back to GitHub + warn, mirroring the existing `Backend` enum pattern), plus host extraction/detection for https, ssh and scp-style remotes.
- `jean.json` gains an optional `provider` block; the Jean.json settings pane and onboarding wizard now preserve fields they don't manage (previously an unrelated edit would silently drop them).
- New `get_git_provider` command + `useProjectGitProvider` hook so the UI can resolve a project's provider.

**GitLab CLI (`glab`) management**
- New `glab_cli` module mirroring `gh_cli`: install/update/uninstall of a Jean-managed binary (from the GitLab releases API), PATH detection, version + auth status, and login via the existing CLI login modal.
- New `glab_cli_source` preference (`jean` | `path`), matching the `gh` equivalent.
- Settings: the **GitHub CLI** pane is renamed **Git Provider** and now hosts both the GitHub CLI and GitLab CLI cards.

**Issues & merge requests**
- New `gitlab_issues` module. It shells out to `glab api` (raw GitLab REST) rather than `glab issue/mr list`, because the REST field names are stable and documented while the subcommands' state/output flags are inconsistent across versions. Host and auth are inferred from the repo directory, the same way `gh api` works.
- Results are mapped into the existing `GitHub*` structs (`iid`→`number`, `description`→`body`, `source/target_branch`→`head/baseRefName`, `draft`→`isDraft`, `author.username`→`login`, state upper-cased), so the UI stays provider-agnostic.
- The existing `list_github_issues` / `list_github_prs` / `get_*` / `search_*` / `list_github_labels` commands now dispatch by provider — no new command names, so MCP and web-access dispatch are untouched.

**Merge request lifecycle**
- `create_pr_with_ai_content` creates a GitLab MR (same AI-generated title/body flow), with an "already exists" fallback that links the existing MR.
- `merge_github_pr`, `link_worktree_pr`, `detect_open_pr_for_branch` and `detect_and_link_pr` are provider-aware.
- `get_pr_status` now dispatches by provider; GitLab MR status maps pipeline status → check status and `(detailed_)merge_status` → mergeable.

**Frontend**
- New Worktree and Load Context modals resolve the provider and gate on the right CLI auth (`glab` vs `gh`), showing a GitLab-specific auth/sign-in prompt.
- Terminology switches to "Merge Request"/"MR" for GitLab projects (tab labels, headings, empty/loading states).
- New GitLab icon; the Security tab is hidden for GitLab projects since Dependabot/advisories are GitHub-only.

## Known limitations / out of scope

- **Security scanning**: GitLab has no equivalent of Dependabot alerts / repository advisories in this integration, so those features are hidden for GitLab projects.
- **Review comments and release notes** are not ported to GitLab yet.
- **Onboarding** still assumes GitHub — there's no provider-selection step, so a GitLab-only user currently sets `glab` up from Settings afterwards. Happy to follow up on this.
- MR reviews/approvals are not mapped onto GitHub's review model (`reviews` is left empty).
- Fork-based push handling (`git_push_to_pr`) is not GitLab-aware; the normal push path works.

## Testing

- `bun run check:all` passes (1406 frontend tests, 803 Rust tests, typecheck, ESLint, rustfmt, clippy).
- Added Rust unit tests for provider detection/parsing (remote URL forms, tolerant enum deserialization, `jean.json` round-trip), `glab` version parsing, and the GitLab→GitHub mapping (state mapping, query params, percent-encoding).
- GitHub behaviour is intended to be unchanged throughout: the provider check short-circuits to the existing `gh` code paths.

## A note on review

I've done a first pass reviewing this myself. Full disclosure on how it was built: I did all of it using **Jean**, partly as a way to see how far I could get dogfooding the tool on a non-trivial feature.

I'm comfortable reading Rust and I understand what the code is doing, but I'm **not a Rust expert** — so I'd really appreciate a second review, particularly on the Rust side (the `glab` process handling, the provider dispatch points, and anything around error handling / edge cases I may have missed).

Close #150


strausmann (commented):
Went through the full diff. Short version: the architecture is the right call, the security-sensitive parts are handled well, and there are a few things I'd want resolved before merge — one a real functional bug in MR creation, and one performance regression that affects GitHub-only projects too.

### Architecture

Dispatching inside the existing commands (rather than adding a parallel GitLab feature) is the right cut: frontend, MCP and HTTP/WS dispatch stay untouched, and the GitHub path short-circuits before any new code runs. Mapping GitLab shapes onto the existing `GitHub*` structs keeps the UI provider-agnostic without a type explosion.

The cost is eight near-identical `if is_gitlab_project(…)` guards in `github_issues.rs`. That's fine for two providers; a third would want a trait.

### What I verified against the live GitLab API

- Asset naming `glab_{version}_{platform}.{ext}` and the download URL `…/-/releases/v{v}/downloads/{asset}` — HTTP 200 for `linux_amd64.tar.gz`, `darwin_arm64.tar.gz` and `windows_amd64.zip` on v1.112.0. The platform strings match the real release assets.
- The releases API is reachable unauthenticated, and `released_at` / `upcoming_release` are mapped correctly.
- Command registration is complete: this repo routes everything through `http_server/dispatch.rs` (no separate `generate_handler!`), and all seven `glab` commands plus `get_git_provider` are wired up there.
- `hasBackend` still exists in `src/lib/environment.ts`, so the new frontend imports won't break on rebase.

### Handled well (worth calling out)

- **No injection surface in the query building.** `pct_encode` encodes everything outside the RFC 3986 unreserved set — including `&`, `=`, `/`, `%` — so a branch/title/body value like `&target_branch=main` can't break out of the query string. And every `glab` call goes through `std::process::Command` with an argument array (no shell), so there's no shell-injection path via branch/title/body either. Nicely done.
- The `create_mr` duplicate fallback (`e.contains("already exists") || e.contains("409")`) matches GitLab's actual response for a duplicate MR (`409 … "This merge request already exists"`), so that path is correct.

### Blockers

**1. `create_mr` sends the whole AI-generated body in the URL query**

```rust
let endpoint = format!(
    "projects/{enc}/merge_requests?source_branch={}&target_branch={}&title={}&description={}",
    …, pct_encode(body),
);
run_glab_api_method(app, project_path, "POST", &endpoint)?
```

AI-generated PR bodies are routinely 1–4 KB, and percent-encoding roughly triples that. GitLab's frontend nginx typically caps the request line around 8 KB, so this will intermittently fail with a 414 — surfaced to the user as a bare `glab api failed: …`. Self-hosted instances are often stricter.

The safe pattern is already in this same file: the GitHub path (`gh pr create --body …`) passes the body as a process argument, never in a URL. `glab api` supports the same via `-f/--raw-field description=…` (string fields, which also flip the method to POST). Moving `create_mr` from a query string to `-f` fields keeps the fix local to `create_mr` + `run_glab_api_method`; the GET endpoints are well under any limit and can stay as they are.

**2. `merge_mr` has no error path and no merge options**

```rust
let endpoint = format!("projects/{enc}/merge_requests/{mr_number}/merge");
run_glab_api_method(app, project_path, "PUT", &endpoint)?;
```

The GitHub path in `merge_github_pr` checks mergeability first and returns actionable errors. GitLab returns 405/409/422 for conflicts, unresolved threads, or "pipeline must succeed" — all of which currently reach the user as raw stderr. `squash` and `should_remove_source_branch` are also unavailable, though the GitHub side has merge options. To be fair this degrades to a raw message rather than a crash (GitLab rejects cleanly server-side), so it's a parity/UX gap rather than a hard failure — but the GitHub path sets the bar here.

**3. Rebase**

The branch is behind `main` with conflicts (`mergeable_state: dirty`), so everything here needs another `bun run check:all` afterwards.

### Performance regression (affects GitHub projects too)

`resolve_git_provider` isn't cached — every call spawns `git remote get-url origin` and reads `jean.json`. I'd initially have filed this as a GitLab-path cost, but on a closer read it's broader: `get_pr_status` now calls `resolve_git_provider` for **every** project — GitHub included — just to pick the branch, and that runs on the background PR-status poller per worktree per cycle. So the PR introduces a new per-poll subprocess spawn for *all* users, not only GitLab ones. A small cache keyed by repo path (or storing the resolved provider on the worktree) removes it and keeps the GitHub path as cheap as before.

### Worth fixing before merge

- **No pagination.** Everything is `per_page=100`, and `total_count = issues.len()`. Past 100 issues the UI silently shows a truncated list plus a wrong total (the GitHub path returns the real `totalCount`).
- **`view_open_mr_for_branch` uses `per_page=1` with no `order_by`/`sort`.** With more than one open MR on the same source branch the pick isn't deterministic — and this is the path that links MRs to worktrees. `order_by=updated_at&sort=desc` would make it predictable.
- **`gitlab_project_path_encoded` only handles the scp form when the remote starts with `git@`.** A remote like `deploy@gitlab.example.com:group/repo.git` errors out, even though `extract_host` in `provider.rs` parses any `user@host:` form fine. The two parsers should share one implementation.
- **No unit test for `gitlab_project_path_encoded`** — it's the one function here with non-trivial URL parsing (scp form, `://` form, trailing-slash / `.git` stripping), and it's exactly where the `git@`-only bug above lives. `provider.rs` already tests `extract_host` across six remote forms; mirroring that here would have caught the `deploy@…` case.
- **Self-hosted GitLab on a host without "gitlab" in the name** (e.g. `git.company.com`) silently falls back to GitHub, and the user then gets confusing `gh` errors rather than a pointer to the `provider` block in `jean.json`. Same shape as #354. A targeted error when the host can't be classified would go a long way.
- **No checksum verification on the `glab` download**, even though every release ships a `checksums.txt`. The `gh` installer doesn't verify either, so this is parity rather than a regression — but it'd be a worthwhile addition for both.

### Minor

- `dist/.gitkeep` is deleted — looks accidental.
- `src-tauri/src/platform/notifications.rs` is a single-line rustfmt reflow unrelated to the feature.

### On closing #150

The description says `Close #150`, but per your own "known limitations" the onboarding provider step, security tab, review comments and fork-push handling aren't covered yet. `Part of #150` seems more accurate so the issue stays open for the remainder.

---

Happy to help land this — I run self-hosted GitLab + `glab` day to day, so the `create_mr`/`-f`, the scp-remote and the merge-error mapping are areas I can contribute concrete fixes for if that's useful. Great groundwork overall.

mgvdev:
Here's a clean markdown block for the PR (English, paste-ready):

---

## Follow-up: addressing review feedback

### 1. What was flagged by review

**Bugs**
- GitLab label chips rendered with an invalid `##` color (fallback color kept a leading `#`).
- Creating a worktree from a Merge Request was broken — it ran `gh pr checkout` against a GitLab repo.

**Risks**
- Several PR actions still shelled out to `gh` on GitLab projects: open-PR-in-browser, link-PR-by-number, push-to-PR, CodeRabbit trigger, PR diff, and PR review comments.
- The "MR already exists" race path returned success without saving the MR link to the worktree (so status/merge couldn't find it).
- Host detection was asymmetric (GitHub Enterprise hosts were "unclassifiable"); local/`file://` origins produced a bogus provider error; a `provider.host` with a scheme (`https://…`) broke every `glab` call.
- Frontend: a GitLab project briefly showed "Sign in to GitHub" while the provider resolved; the Security tab was hidden but still reachable via keyboard; caches weren't invalidated after `glab` login or after saving a provider to `jean.json`; a save race could drop the `provider` block.
- New provider-aware behavior had no tests.

**Nits**
- Dead code (`resolve_gitlab_api_token`), GitLab `locked` state fell through, provider-guard applied inconsistently, and unrelated rustfmt churn in non-feature files.

### 2. What I changed and why

- **Made every git-host operation provider-aware.** Issues/MRs already dispatched by provider; I extended the same dispatch to checkout (fetch `refs/merge-requests/<iid>/head`), diff (`glab mr diff`), review comments (mapped from MR discussions), open-in-browser (`glab mr create --web`), link/detect, and push — so GitLab projects no longer hit `gh` and fail.
- **Fixed the correctness bugs**: bare-hex label color; persist the MR link on both create-success and the already-exists race so downstream status/merge work; symmetric host detection; ignore hostless origins; normalize `provider.host` before passing it to `glab`; map the `locked` MR state; apply the unconfigured-provider guard to all GitHub commands.
- **Frontend polish**: defer the auth-error UI until the provider is known (no GitHub flash on GitLab), invalidate the right query caches after login/reinstall and after saving `jean.json`, fully hide the Security tab (content + keyboard shortcut) for GitLab, and block Save until the config has loaded so the `provider` block can't be dropped. Added tests covering the GitLab paths (GitLab auth error, "MR" wording, config preservation).
- **Removed the unrelated formatting churn**: it came from a local rustfmt version reflowing files it shouldn't have. Reverted those non-feature files to `main` (verified they were formatting-only). Note: CI doesn't run a `cargo fmt` check, so this was purely to keep the diff focused.
- CodeRabbit-on-GitLab now returns a clear "not supported yet" message instead of a raw `gh` error.

---

strausmann:
One more from testing this against a self-hosted GitLab 19.1.0-ee instance (nice turnaround on the three blockers, by the way):

`merge_mr` fails there with `400 SHA must be provided when merging`. That instance requires the `sha` param on the merge endpoint — the mergeability pre-check and the squash / remove-source-branch resolution both work correctly; it's only the merge call itself that's rejected. Merging the same MR by hand with `sha` supplied succeeds (`state: merged`, honoring the project's `squash_option` / `remove_source_branch_after_merge`).

Since `fetch_mr_status_raw` already loads the MR, the fix should be small: add `diff_refs.head_sha` (or `sha`) to `GlMrStatus` and pass it as a third `--raw-field sha=…` on the merge call. Passing the SHA is a harmless safety check on instances that don't require it, so it shouldn't regress the ones that don't.

PR #518, opened 2026-07-21
fix(chat): make WSL Claude sessions survivable

CamAnNguyen:
## Problem

Starting a Claude session from Jean on Windows with WSL mode enabled would hang for 120s and then die silently. No output, no error in the chat — just a spinner until the startup timeout, and a `wsl.exe` left pinned for the whole run.

Two independent bugs stacked on top of each other, which is why it looked like a hang instead of a failure.

### 1. The Linux child was killed the moment `wsl.exe` exited

Jean launched Claude with `wsl.exe -- sh -c "... nohup claude ... & echo $!"`. That backgrounded child stays in the launcher's session. When the launcher finishes and `wsl.exe` tears the session down, the child gets SIGHUP'd and dies — instantly, before it ever writes a byte to the output file.

Windows *does* have a handoff for this: when `wsl.exe` exits before the Linux process finishes, `wslhost.exe` [takes over the lifetime of the Linux process](https://github.com/microsoft/WSL/blob/master/doc/docs/technical-documentation/wslhost.exe.md). The catch neither doc spells out is that adoption only saves a process that is *still alive* when `wsl.exe` leaves. A plain backgrounded child sharing the launcher's session is reaped at teardown — before adoption ever applies. `nohup` doesn't help either; it covers the `nohup`'d command, not the session teardown that kills it.

### 2. pid 0 reads as "alive" forever

The spawn path took whatever `echo $!` printed and trusted it. If the session never came up — a `--cd` into a directory that doesn't exist inside the distro, or a pid mangled by Windows→WSL argument rewriting — that value came back as `0`.

`is_process_alive(0)` in WSL mode shells out to `kill -0 0`. Signal 0 with pid 0 targets **the caller's process group**, not "process zero" — so it always succeeds. A dead spawn reported alive, forever. The tailer then sat on an empty output file waiting for a process that never existed, until the 120s startup timeout, and even then exited quietly with no `chat:error`.

A third contributor fed bug 2: `win_to_wsl_path` didn't strip Windows verbatim prefixes. A canonicalized cwd comes back as `\\?\C:\..` or `\\?\UNC\wsl.localhost\..`, fell through untranslated, and produced a `--cd` that couldn't resolve inside the distro — chdir fails → pid 0 → silent hang.

## Fix

**Launch Claude in its own session.** The launcher now runs `setsid sh -c '...' </dev/null &`, giving Claude a new session with no controlling terminal. It survives the launcher's teardown, `wslhost.exe` adopts it, and `wsl.exe` returns immediately instead of being pinned for the entire run. Verified on WSL2: the child stays alive with its own SID after the launcher is gone.

**Report Claude's real pid.** The inner session does `echo $$ > pidfile; exec claude ...`, so the pid Jean records is Claude's own stable Linux pid. Reading `$!` of `setsid` would give the short-lived forking parent — a pid that is dead or zero by the time anyone checks it. The outer launcher waits (≤2s) for the pid file to land, prints it, and exits. The pid file is deleted right after it's read; only the run path ever touches disk — the MCP token stays in Claude's argv/env.

**Feed the script over stdin.** `sh -s` with the program written to stdin, instead of `sh -c "<program>"` as an argv element, so Windows→WSL argument rewriting can't mangle it.

**Make pid 0 impossible to mistake for a live process:**
- `is_process_alive(0)` returns `false` on both Unix and Windows, before the `kill -0` / `OpenProcess` call.
- Every detached spawn branch (Unix, WSL, native Windows) rejects pid 0 with a real error instead of handing it to recovery state.

**Fail loudly on startup timeout.** If Claude produced no output before the timeout, Jean now emits a `chat:error` with the CLI's stderr when there is any, or an actionable message pointing at the Claude CLI install / distro config when there isn't — and terminates the process tree so nothing is left running behind a dead session.

**Path translation fixes:**
- `win_to_wsl_path` strips `\\?\` and `\\?\UNC\` prefixes, so a canonicalized cwd resolves inside the distro.
- Path-valued flags (`--add-dir`, `--append-system-prompt-file`, `--settings`) get their Windows values translated to WSL paths. Only values that actually *look* like Windows paths (`C:\..`, `\\..`) are touched, so inline JSON for `--settings` and non-path values like `--model` pass through untouched. `--append-system-prompt-file` was the worst of these — a Windows path there aborts the entire run.

## Changed files

| File | What |
|---|---|
| `jean-core/src/chat/detached.rs` | `setsid` launcher, pid-file capture, stdin-fed script, pid 0 rejection, path-flag translation |
| `jean-core/src/chat/claude.rs` | Terminal `chat:error` + process-tree kill on startup timeout |
| `jean-core/src/platform/process.rs` | `is_process_alive(0) == false` on Unix and Windows |
| `jean-core/src/platform/wsl.rs` | Strip `\\?\` / `\\?\UNC\` verbatim prefixes |

## Testing

- WSL2 (Ubuntu), Windows host: session starts, streams, and completes; `wsl.exe` exits immediately after spawn while Claude keeps running with its own SID.
- Unit tests added for `wslify_path_args` (path flags translated, non-path values and inline `--settings` JSON untouched), `startup_failure_message` (uses CLI stderr when present, actionable fallback when not, silent on user cancel / completed runs), `is_process_alive(0)`, and the new `win_to_wsl_path` prefix cases.
- Non-WSL paths unchanged: Unix and native Windows spawn only gained the pid 0 guard.

Main changes from current body: separated the two bugs with why each one bites (session teardown vs kill -0 0 targeting the process group), added the wslhost.exe adoption nuance that explains why nohup alone was never enough, folded the verbatim-prefix bug in as a cause of pid 0 instead of a loose bullet, and expanded the "additional improvements" one-liner into concrete items with rationale.

### DIFF
--- a/jean-core/src/chat/claude.rs
+++ b/jean-core/src/chat/claude.rs
@@ -1298,6 +1298,40 @@ fn flush_pending_chunks(
     }
 }
 
+fn startup_failure_message(
+    startup_failed: bool,
+    user_cancelled: bool,
+    full_content: &str,
+    error_lines: &[String],
+) -> Option<String> {
+    if !startup_failed || user_cancelled || !full_content.is_empty() {
+        return None;
+    }
+
+    if !error_lines.is_empty() {
+        return Some(format!("Claude CLI failed: {}", error_lines.join("\n")));
+    }
+
+    Some(
+        "Claude CLI produced no output and was stopped. It may have failed to start; \
+         check the Claude CLI installation and, in WSL mode, the distro configuration."
+            .to_string(),
+    )
+}
+
+fn terminate_failed_startup(pid: u32) {
+    if pid <= 1 {
+        return;
+    }
+
+    if let Err(tree_error) = crate::platform::kill_process_tree(pid) {
+        log::warn!("Failed to terminate startup process group {pid}: {tree_error}");
+        if let Err(process_error) = crate::platform::kill_process(pid) {
+            log::warn!("Failed to terminate startup process {pid}: {process_error}");
+        }
+    }
+}
+
 /// Tail an NDJSON output file and emit events as new lines appear.
 ///
 /// This is used for detached Claude CLI processes where the CLI writes
@@ -1344,6 +1378,7 @@ pub fn tail_claude_output(
     let mut completed = false;
     let mut cancelled = false;
     let mut user_cancelled = false; // True only for explicit user cancel (not process death)
+    let mut startup_failed = false; // True when Claude produced no output before the startup timeout / died starting up
     let mut usage: Option<UsageData> = None;
     let mut error_lines: Vec<String> = Vec::new();
 
@@ -2344,6 +2379,7 @@ pub fn tail_claude_output(
                     elapsed.as_secs_f64()
                 );
                 cancelled = true;
+                startup_failed = true;
                 break;
             }
 
@@ -2353,6 +2389,7 @@ pub fn tail_claude_output(
                     startup_timeout
                 );
                 cancelled = true;
+                startup_failed = true;
                 break;
             }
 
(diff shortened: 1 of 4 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #529, opened 2026-07-22
feat: Add agent prompt copy button

### LINKED ISSUE #527: No agent response copy button
A copy button for user prompts exists, but not for agent returns apparently.

alepouna:
Adds a copy-to-clipboard button at the bottom-right of completed agent responses in the chat message view.

Fixes #527

### DIFF
--- a/src/components/chat/MessageItem.test.tsx
+++ b/src/components/chat/MessageItem.test.tsx
@@ -1,5 +1,5 @@
 import { describe, expect, it, vi } from 'vitest'
-import { render, screen } from '@/test/test-utils'
+import { render, screen, waitFor } from '@/test/test-utils'
 import { MessageItem } from './MessageItem'
 import type {
   ChatMessage,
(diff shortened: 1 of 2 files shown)


andrasbacsai (dismissed):
## Review

Looks good — this is the right fix for #527.

### What works well
- Copy affordance only on completed assistant `MessageItem` rows (streaming path correctly left alone).
- Prefers `message.content`, falls back to ordered text `content_blocks` — covers the common empty-`content` + blocks case.
- Hover-reveal styling matches the existing user-message copy control, including coarse-pointer visibility.
- Bottom-right placement avoids overlapping long replies.
- Uses the shared `copyToClipboard` helper + success/error toasts.
- Solid tests for content and content-blocks paths.

### Nits (non-blocking)
1. **Toast copy length** — user prompts toast `Prompt copied`; this uses the longer `Response copied to clipboard`. Prefer matching the shorter style (`Response copied`) for consistency.
2. **Empty text blocks** — `flatMap(block => (block.type === 'text' ? [block.text] : []))` includes blank `text` blocks in the join (final `.trim()` only cleans ends). Filtering `block.text.trim()` before join is slightly cleaner.
3. **Tests** — nice-to-have: assert the button is hidden when there is no copyable text, and that failure shows the error toast (implementation already handles both).
4. **PR title** — says “agent **prompt** copy button”; this copies the **response**. Minor naming nit only.

None of these block merge. Thanks for picking this up!

**Verdict:** Approve

andrasbacsai:
Thank you for the PR! 💜

PR #562, opened 2026-07-23
feat(worktrees): create a worktree from any remote's default branch

fsioni:
## Why

A bit of context: I run Jean on a fork of Jean itself — `origin` is `coollabsio/jean`, `fork` is mine. Managing that fork and sending improvements back here is honestly my first use case for the app, so this is the first paper cut I ran into. 😄

Today the New Worktree modal can only branch off the project default branch as it exists locally. Whichever remote my local `main` happens to track silently decides where a new worktree starts — there is no way to say "start this one from `origin/main`" while my local `main` follows `fork/main`, or the other way round. I end up leaving the app and doing it by hand.

## What

**Actions tab** — when a project has more than one remote carrying the default branch, the tab shows one action per remote (origin first) instead of the single generic "New Worktree" tile, sharing one *Branch name (optional)* field:

New Worktree actions with one tile per remote

- Projects with a single matching remote are untouched: same "New Base Session" / "New Worktree" tiles as before.
- Remotes that don't have the branch fetched are filtered out, so an action is never offered for a start point that cannot resolve.
- The base session stays reachable as a discreet link under the tiles. `N` and `Enter` create from the first remote, the one carrying the `N` badge.

**Backend** — `create_worktree` now accepts a remote-qualified base such as `fork/main`: it fetches that remote first, then uses `<remote>/<branch>` as the git start point. Previously such a value silently fell back to the local branch. A qualified base that isn't fetched locally now returns an explicit error instead of quietly starting somewhere else.

Ordinary slashed branch names (`feature/login`) and PR heads already tracked on origin keep taking the historical origin-based path — both are covered by tests.

**Afterwards** — the picked remote is stored on the worktree and surfaces in the existing base-branch badge, in the session header and the project canvas, so a worktree started from `fork/main` says so instead of showing a bare `main` (which is hidden altogether today when the base equals the default branch):

Session header showing the fork/main base badge

**Status comparisons follow the same remote.** Git status hardcoded `origin/<base branch>`, so a worktree started from `fork/main` counted every commit that remote is ahead of origin as its own work — a brand new session showed `+19431/-313` and 33 unpushed commits here. Ahead/behind counts, branch diff stats and the diff modal now compare against `<remote>/<base branch>` and fetch that remote for freshness. Worktrees without an explicit base remote keep the previous origin-based behaviour, and the worktree still stores the short branch name, so nothing changes for single-remote projects.

The new `list_remotes_with_branch` command is registered in the core dispatch, so it works over web access too.

## Tests

- 10 Rust unit tests: remote resolution (`remotes_with_branch`, `remote_tracking_branch_exists`, qualified-base splitting) in `jean-core/src/projects/git.rs`, plus `base_ref` in `git_status.rs`.
- 12 front tests: 9 in `src/components/worktree/QuickActionsTab.test.tsx` covering both layouts, the shared branch name, the keyboard paths and invalid names, plus 3 in `worktree-branch-badge.test.ts` for the remote-qualified badge.
- `bun run typecheck`, `bun run rust:clippy` and `bun run rust:test` (835 tests) are clean. `bun run test:run` passes 1438 tests, the only failing file being `src/components/chat/toolbar/MobileSettingsMenu.test.tsx`, which already fails on `main` (a `vi.mock` hoisting error). `bun run lint` reports 4 pre-existing errors in that same file and `RemoteConnectionsDialog.test.tsx`. None of those files are touched here.

Happy to adjust the UI if you'd rather keep the Actions tab to a single tile — the backend part stands on its own.


andrasbacsai:
Thank you for the PR! 💜

PR #567, opened 2026-07-23
fix(markdown): preserve double-digit ordered list markers

### LINKED ISSUE #542: Bug: Number list doesn't properly show number over 10
Image

After 9 it goes like `.0`, `.1` .... instead of `10`, `11`... 

Jean Version: 0.1.68
OS: MacOS Sequoia

horacioh:
## Summary

- widen the shared Markdown ordered-list gutter so double-digit markers remain visible
- preserve outside markers and wrapped-line alignment
- add regression coverage for chat, compact, streaming, tool-call, and streaming tool-call Markdown modes

## Root cause

The Markdown renderer used outside decimal markers with a `pl-6` gutter while chat surfaces clip horizontal overflow. Starting at item 10, the leading digit extended beyond the visible gutter and was clipped.

## Testing

- `bun x vitest run src/components/ui/markdown.test.tsx` — 11 tests passed
- changed-file ESLint, Prettier, and `git diff --check`
- browser smoke test with 12 items in chat and compact-preview surfaces; confirmed complete markers, aligned wrapping, and no horizontal overflow

Closes #542

### DIFF
--- a/src/components/ui/markdown.test.tsx
+++ b/src/components/ui/markdown.test.tsx
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest'
-import { render, screen } from '@/test/test-utils'
+import { render } from '@/test/test-utils'
 import { Markdown } from './markdown'
 
 describe('Markdown', () => {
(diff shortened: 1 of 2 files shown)


horacioh:
closing this in favor of #568

PR #570, opened 2026-07-25
fix(docker): install GitHub CLI in server images

midego1:
The onboarding flow has a required **"Authenticate GitHub CLI"** step that shells out to `gh`, but the server runtime image ships only `ca-certificates`, `curl`, `git` and `openssh-client`. In a container the step fails immediately with:

```
[Process exited with code unknown]
Login process exited unexpectedly — Exit code: unknown
```

which doesn't hint that the binary is missing.

**Repro** — `ghcr.io/coollabsio/jean-server:v0.1.68`, headless (`JEAN_HEADLESS=1`):

```
$ docker exec <container> sh -c 'command -v gh || echo missing'
missing
```

**Change** — install `gh` from the official apt repository in the runtime stage of both server Dockerfiles, inside the existing layer so no extra layer is added. Verified on `debian:bookworm-slim`: `gh version 2.96.0`, `jean` user unchanged.

**Note** — with `gh` present the prompts render and arrow-key selection works, but I still couldn't finish the login from the dialog: after authorizing the device code on github.com there was no way to advance, and `gh` never wrote `~/.config/gh/hosts.yml`. Running `gh auth login --hostname github.com --git-protocol https --web < /dev/null` via `docker exec` worked, and `gh auth setup-git` was needed afterwards or `git push` still prompts. That looks separate from this change — happy to file it as its own issue if useful.

### DIFF
--- a/Dockerfile.server
+++ b/Dockerfile.server
@@ -22,6 +22,12 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     curl \
     git \
     openssh-client \
+    && curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg \
+        -o /usr/share/keyrings/githubcli-archive-keyring.gpg \
+    && chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg \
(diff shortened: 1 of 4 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #602, opened 2026-07-29
fix(focus): restore keyboard focus after alt-tab

### LINKED ISSUE #577: Bug: Alt-tabbing on Jean loses focus of prompt entry
If you tab out, change windows, etc. when re-entering Jean you have to click anywhere on Jean. Ctrl +L (cmd + L) refuses to work until Jean is interacted with by the mouse.

andrasbacsai:
## Summary
- Restore keyboard focus when Jean is re-activated after alt-tab or window switch, so the chat prompt and shortcuts like Ctrl/Cmd+L work without a mouse click
- Prefer the last focused control (or chat input), and avoid stealing focus from terminals, dialogs, menus, or other intentional inputs
- Add unit coverage for restore paths and chat window re-focus behavior

fixes #577

---

Fixes #577

### DIFF
--- a/src/components/chat/ChatInput.tsx
+++ b/src/components/chat/ChatInput.tsx
@@ -1215,6 +1215,7 @@ export const ChatInput = memo(function ChatInput({
       />
       <Textarea
         ref={inputRef}
+        data-chat-input
         placeholder={
           isSending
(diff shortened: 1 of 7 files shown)


andrasbacsai:
@coderabbitai review

PR #603, opened 2026-07-29
fix(grok): surface rate limits and ACP errors

### LINKED ISSUE #580: Bug: Grok does not handle limit or other error responses
Image

andrasbacsai:
## Summary

- Fix Grok turns that failed with rate limits, auth errors, or JSON-RPC prompt failures completing as empty runs instead of showing an error
- Write a single host terminal marker (`error` or `result`, never both) so failed ACP prompts are not salvaged as successful completions
- Parse and format user-facing messages for rate/usage limits and authentication failures, and surface them via `chat:error`
- Prefer error markers over result lines when recovering run history so legacy dual-marker logs no longer produce “content was not captured”

fixes #580

---

Fixes #580

### DIFF
--- a/jean-core/src/chat/codex.rs
+++ b/jean-core/src/chat/codex.rs
@@ -735,10 +735,7 @@ pub fn apply_codex_provider_to_config(
         provider_entry.insert("wire_api".to_string(), serde_json::json!(wire));
     }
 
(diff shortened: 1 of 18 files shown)


andrasbacsai:
@coderabbitai review

PR #604, opened 2026-07-29
fix(opencode): surface child errors and stop POST retries

### LINKED ISSUE #587: Bug: OpenCode provider errors can still be dropped for child sessions and timed-out prompts are retried
## Summary

When an OpenCode provider rejects a prompt, Jean can mask the real provider error behind a long-running request and can submit the same prompt twice.

I reproduced this with the Windows native Jean v0.1.68 release, OpenCode CLI 1.18.8, and `opencode-go/kimi-k3` after its five-hour allowance was exhausted. OpenCode logged the actionable quota error about 300 ms after the request, but Jean remained in the sending state, retried the message after the 1,800-second HTTP timeout, and eventually showed only:

```text
Failed to send OpenCode message: error sending request for url
(http://127.0.0.1:4096/session/<session-id>/message?directory=C%3A%5CUsers%5C...)
```

The percent-encoded Windows `directory` query is valid. The URL/path encoding is not the cause.

[#523](https://github.com/coollabsio/jean/issues/523) and merged [#539](https://github.com/coollabsio/jean/pull/539) fix the primary/direct-session SSE error path on current `main`. This issue tracks gaps that remain after that fix: child/subagent errors are not associated with the parent subscriber, message POSTs are still retried after a request-level failure or timeout, and the managed OpenCode server discards its diagnostics.

## Environment

- OS: Windows 11 Pro 24H2, build 26100, 64-bit
- Installed Jean: v0.1.68 (latest published release at time of reproduction)
- OpenCode CLI: 1.18.8
- Jean backend: OpenCode
- Provider/model: `opencode-go/kimi-k3`
- OpenCode authentication: configured and working
- Current `main` inspected at: `e19c544e42e6519f08b0815de3b63a348a8c4e39` (`package.json` reports 0.1.69, but no v0.1.69 release is published yet)

### Preconditions

1. Install Jean v0.1.68 on Windows.
2. Install OpenCode CLI 1.18.8 and authenticate the OpenCode Go provider.
3. Exhaust the model's five-hour allowance, or use a provider/model that emits `session.error` while the `/message` response remains open.
4. Add any local Git repository to Jean and open or create a worktree/session.

### Steps

1. Start the native Jean desktop app.
2. Open the project and an active worktree.
3. In the chat toolbar, select the **OpenCode** backend.
4. Select provider/model **OpenCode Go / Kimi K3** (`opencode-go/kimi-k3`).
5. Send a simple prompt such as:

   ```text
   Reply with the current branch name.
   ```

6. Immediately inspect OpenCode's Windows log:

   ```powershell
   Get-Content "$env:USERPROFILE\.local\share\opencode\log\opencode.log" -Wait |
     Select-String 'stream error|usage limit|session.error'
   ```

7. Observe that OpenCode logs the real error almost immediately. In the reproduced session it was:

   ```text
   AI_APICallError: 5-hour usage limit reached.
   ```

8. Leave Jean running. Jean stays in the sending/running state instead of displaying that error.
9. After approximately 1,800 seconds, inspect/export the OpenCode session. The same user prompt is present a second time because Jean retries the POST.
10. After the second long wait, Jean reports the generic `Failed to send OpenCode message: error sending request for url (...)` error instead of the provider error.

### Observed timing from the reproduced session

```text
15:50:32.456 OpenCode session created
15:50:33.547 provider stream started (opencode-go/kimi-k3)
15:50:33.852 OpenCode logged AI_APICallError: 5-hour usage limit reached
~1,800 seconds later Jean submitted the user message again
~1,800 seconds after that Jean surfaced the generic request URL error
```

## Reproduction 2: current `main` can ignore child/subagent provider errors

This exercises the gap remaining after #539.

### Preconditions

1. Build current Jean `main` on Windows following [CONTRIBUTING.md](https://github.com/coollabsio/jean/blob/main/CONTRIBUTING.md):

   ```powershell
   bun install
   bun run tauri:dev
   ```

2. Configure an OpenCode model that will immediately return a provider error, such as an exhausted OpenCode Go quota.
3. Open a Jean session using OpenCode in Build or YOLO mode.

### Steps

1. Send a prompt that explicitly causes OpenCode to use child agents, for example:

   ```text
   Use multiple explore subagents to inspect this repository, then summarize their results.
   ```

2. Tail the OpenCode log with the command above.
3. Confirm that OpenCode emits `stream error` / `session.error` entries for child session IDs (`mode=subagent`, commonly `agent=explore`).
4. Observe the corresponding Jean parent session.

### Actual result

The SSE subscriber registry is keyed by the primary OpenCode session ID. `surface_opencode_sse_error()` returns `false` when the error's child session ID is not present in that registry, so the parent Jean run may continue to wait or fail later without the child's actionable provider error.

In the observed log, multiple `mode=subagent` sessions reported the five-hour usage-limit error while the Jean run was active.

### Expected result

- A provider failure from the primary OpenCode session **or any descendant child session belonging to the run** should immediately fail the owning Jean run with the provider's actionable message.
- Jean must submit a user prompt at most once. A timed-out or ambiguous non-idempotent `/message` POST must not be automatically replayed.
- The UI should not classify a provider failure as a user cancellation.
- Jean diagnostics should retain enough sanitized OpenCode server output to explain startup/provider failures.

## Why #523 / #539 do not fully cover this

- [#523](https://github.com/coollabsio/jean/issues/523) reported local/custom providers hanging.
- [#539](https://github.com/coollabsio/jean/pull/539) correctly added handling for primary-session `session.error` and `message.updated` events and unblocks the POST with `force_abort`.
- The current handler still looks up the exact event `sessionID` in the subscriber map and returns without surfacing the error when it is a child ID.
- `post_opencode_message_cancellable()` still retries when `e.is_connect() || e.is_request()`. A request-level error includes ambiguous cases where the server may already have accepted the prompt; replaying the POST is unsafe.
- The HTTP timeout remains 1,800 seconds.
- Managed OpenCode stdout and stderr are still sent to `Stdio::null()`.
- There is no integration test covering an SSE error arriving while the message POST remains blocked, nor a test asserting only one POST after a timeout.

Related but separate: [#580](https://github.com/coollabsio/jean/issues/580) reports unhandled limit/error responses for Grok. That suggests a similar product-level expectation, but it is a different backend path and does not address OpenCode session routing or retry safety.

## Relevant code

- Primary SSE error lookup and early return for unknown session IDs: [`jean-core/src/chat/opencode.rs#L1131-L1146`](https://github.com/coollabsio/jean/blob/e19c544e42e6519f08b0815de3b63a348a8c4e39/jean-core/src/chat/opencode.rs#L1131-L1146)
- `message.updated` / `session.error` routing: [`opencode.rs#L2096-L2140`](https://github.com/coollabsio/jean/blob/e19c544e42e6519f08b0815de3b63a348a8c4e39/jean-core/src/chat/opencode.rs#L2096-L2140)
- 1,800-second timeout and broad retry: [`opencode.rs#L2233-L2285`](https://github.com/coollabsio/jean/blob/e19c544e42e6519f08b0815de3b63a348a8c4e39/jean-core/src/chat/opencode.rs#L2233-L2285)
- Provider-error/POST race added by #539: [`opencode.rs#L2435-L2549`](https://github.com/coollabsio/jean/blob/e19c544e42e6519f08b0815de3b63a348a8c4e39/jean-core/src/chat/opencode.rs#L2435-L2549)
- Managed server output discarded: [`jean-core/src/opencode_server/mod.rs#L201-L208`](https://github.com/coollabsio/jean/blob/e19c544e42e6519f08b0815de3b63a348a8c4e39/jean-core/src/opencode_server/mod.rs#L201-L208)

## Acceptance criteria

- [ ] Primary and descendant OpenCode `session.error` / `message.updated.info.error` events fail the owning Jean run immediately with the provider message.
- [ ] A non-idempotent OpenCode message POST is never automatically repeated after a timeout or ambiguous request error.
- [ ] Exactly one user message appears in the OpenCode session for one Jean send action.
- [ ] Provider errors remain distinct from explicit user cancellation.
- [ ] Add a regression test for a child-session error mapped to its parent run.
- [ ] Add a regression/integration test where an SSE provider error arrives while the POST body remains blocked.
- [ ] Add a regression test proving a timed-out POST is attempted once.
- [ ] Preserve sanitized managed-server diagnostics rather than discarding all stdout/stderr.

andrasbacsai:
## Summary

- Surface OpenCode provider errors from primary **and** descendant/child (subagent) sessions so the owning Jean run fails immediately with the actionable provider message instead of hanging
- Stop automatic retries of non-idempotent OpenCode `/message` POSTs after timeouts or ambiguous request-level failures (one Jean send → one user message)
- Keep provider failures distinct from explicit user cancellation
- Retain sanitized managed OpenCode server stdout/stderr diagnostics instead of discarding them
- Add regression coverage for child-session error routing, SSE error while POST is blocked, single timed-out POST attempt, and diagnostic sanitization/redaction

## Breaking Changes

None.

fixes #587

---

Fixes #587

### DIFF
--- a/jean-core/src/chat/opencode.rs
+++ b/jean-core/src/chat/opencode.rs
@@ -5,7 +5,7 @@ use crate::http_server::EmitExt;
 use base64::{engine::general_purpose::STANDARD, Engine};
 use once_cell::sync::Lazy;
 use regex::Regex;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::future::Future;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
@@ -342,6 +342,10 @@ fn clear_listener_cooldown(state: &SharedSseListenerState) {
 struct SharedSseCoordinator {
     subscribers: Arc<Mutex<HashMap<String, SharedSseSubscriberEntry>>>,
     listeners: Arc<Mutex<HashMap<String, SharedSseListenerState>>>,
+    /// Maps descendant OpenCode session IDs (subagents/children) to a known
+    /// ancestor ID. Walked until a registered primary-session subscriber is found
+    /// so child `session.error` events fail the owning Jean run.
+    session_parents: Arc<Mutex<HashMap<String, String>>>,
 }
 
 impl Default for SharedSseCoordinator {
@@ -349,6 +353,7 @@ impl Default for SharedSseCoordinator {
         Self {
             subscribers: Arc::new(Mutex::new(HashMap::new())),
             listeners: Arc::new(Mutex::new(HashMap::new())),
+            session_parents: Arc::new(Mutex::new(HashMap::new())),
         }
     }
 }
@@ -461,6 +466,12 @@ impl Drop for SharedSseSubscription {
                 subscribers.len()
             );
         }
+        // Drop parent-chain entries that point at (or are) this root so a later
+        // session never inherits a stale child→parent mapping.
+        let mut parents = lock_recover(&SHARED_SSE.session_parents, "OPENCODE_SSE_SESSION_PARENTS");
+        parents.retain(|child, parent| {
+            child != &self.opencode_session_id && parent != &self.opencode_session_id
+        });
     }
 }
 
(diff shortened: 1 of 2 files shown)


andrasbacsai:
@coderabbitai review

PR #605, opened 2026-07-29
feat(chat): retain scroll position when switching sessions

### LINKED ISSUE #594: Improvement: Retain scroll position when switching sessions
Currently changing the currently viewed session always scrolls you back to the bottom. Would be nice if the scroll position was remembered, at least while Jean is open.

andrasbacsai:
## Summary
- Remember each session’s scroll position and expanded history window while Jean stays open, so switching sessions no longer always jumps to the bottom
- Restore scroll against the actually mounted transcript (deferred session + content-ready), and clear saved state when a session is closed, archived, or history is cleared
- Add coverage for save/restore, readiness guards, and session lifecycle cleanup

fixes #594

---

Fixes #594

### DIFF
--- a/src/components/chat/ChatWindow.tsx
+++ b/src/components/chat/ChatWindow.tsx
@@ -1248,7 +1248,11 @@ export function ChatWindow({
   } = useScrollManagement({
     messages: session?.messages,
     virtualizedListRef,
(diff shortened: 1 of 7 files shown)


andrasbacsai:
@coderabbitai review

PR #619, opened 2026-07-31
fix(ui): prevent display-scale zoom feedback loops

fsioni:
## Summary
- listen only to Tauri's native display-scale event when refreshing custom desktop zoom
- avoid reapplying 100% zoom and guard custom zoom refreshes against re-entrant scale events
- remove the resize/devicePixelRatio fallback that can feed WebView zoom changes back into the layout

## Why
The display-scale zoom refresh introduced in `7becad2b` can repeatedly call `setZoom()` when the resulting WebView resize or DPR update is observed as another scale change. At 100% this makes the whole interface continuously jump until the user changes zoom.

## Test plan
- [x] `bun run test:run src/hooks/use-zoom.test.tsx`
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] Manually verify 100% zoom remains stable while resizing and moving between displays
- [x] Manually verify a custom zoom is preserved when moving between displays

### DIFF
--- a/docs/developer/troubleshooting.md
+++ b/docs/developer/troubleshooting.md
@@ -219,7 +219,8 @@ defaults -currentHost write -g AppleFontSmoothing -int 2
 **Engineering notes:**
 
 - Native zoom is applied via `getCurrentWebview().setZoom()` in `src/hooks/use-zoom.ts`.
-- On display scale changes, Jean re-applies zoom (bounce through 1.0 when target ≠ 100%) so WKWebView refreshes its layer after multi-monitor moves.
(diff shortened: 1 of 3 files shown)


fsioni:
Really annoying bug. I think we should fix it before next release
[Capture vidéo du 2026-07-31 15-52-02.webm](https://github.com/user-attachments/assets/74b5da20-f7f3-4866-b7e1-cf18eae0164e)

fsioni:
up @andrasbacsai 
Latest release shipped the bug to users. It makes the app unusable. Can you check? Thanks

andrasbacsai:
Thank you for the PR! 💜

PR #650, opened 2026-08-05
fix(projects): refresh replaced avatar images

fsioni:
## Fixes
Replacing an avatar image didn't work, even after removing the old one

## Summary
- store each imported project avatar under a fresh UUID-based filename
- remove legacy and previous avatar files before copying the replacement
- add frontend and Rust regression coverage for avatar replacement caching

## Testing
- `bun --bun node_modules/vitest/vitest.mjs run src/components/projects/panes/GeneralPane.test.tsx`
- `cargo test --manifest-path jean-core/Cargo.toml project_avatar_replacements_get_a_fresh_file_name`
- `bun run typecheck`
- `bun run lint`


andrasbacsai:
Thank you for the PR! 💜

PR #656, opened 2026-08-07
fix(files): open files in view mode by default

manstfu:
Clicking a file in the file browser opened it directly in the Pierre editor, which makes accidental edits easy and adds an editor load for what is usually just a read.

`FileContentModal` now leaves `isEditing` at its initial `false` after loading content, so the file renders in the syntax-highlighted read-only view. The `Edit` button in the modal header switches to the editor exactly as before, and saving, discarding, and "Open in Editor" are unchanged.

This applies everywhere the modal is used: the desktop file browser sidebar, the mobile file browser, and file links in tool calls.


andrasbacsai:
Thank you for the PR! 💜

PR #659, opened 2026-08-07
fix(chat): recover completed runs before duplicate guard

fsioni:
## Summary

image

- reconcile runs that are still marked `Running` when their JSONL journal already contains a final `result`
- perform this recovery before rejecting a new run as a duplicate
- clear stale review state when a completed run is recovered

## Problem

If Jean was interrupted after writing the final JSONL `result` but before persisting the completed run metadata, the session remained marked as running. Every subsequent message was rejected with:

> Session already has a Running run — refusing to create duplicate

The conversation itself was no longer active, so users could not resume it.

## Testing

- Added regression test: `completed_jsonl_run_does_not_block_the_next_run`
- `cargo test chat::run_log::tests::` — 21 passed
- `rustfmt --check --edition 2021 jean-core/src/chat/run_log.rs`
- Built and manually tested an AppImage against the affected real session; sending a new message succeeded without modifying or resetting the existing configuration


andrasbacsai:
Thank you for the PR! 💜

PR #672, opened 2026-08-11
feat(worktrees): add prompt-first creation flow

fsioni:
## Summary

- add the prompt-first worktree/session creation flow
- preserve the complete queued message while setup runs so backend, model, mode, and attachments survive reloads
- resume pending prompts after setup completes without requiring the chat window to mount
- discard orphaned recovery state when its worktree no longer exists, preventing an infinite retry/toast loop
- improve the related mobile flow, setup progress UI, and recovery coverage

## Bug fixed

A persisted setup prompt could reference a worktree that had since been deleted. Recovery treated `Worktree not found` as transient, retried every second, and displayed an error toast on every attempt.

The recovery hook now clears the stale prompt/message state for this terminal error and does not schedule another retry.

## Screenshot

Repeated pending-prompt recovery errors

## Testing

- `usePendingSetupMessageRecovery.test.ts`: 4 tests passed
- `ChatInput.test.tsx`: 21 tests passed
- TypeScript typecheck passed
- ESLint passed for the recovery implementation and regression test

> Note: linting the upstream-modified `ChatInput.test.tsx` currently reports the pre-existing `@typescript-eslint/consistent-type-imports` violation on its `importOriginal<typeof import(...)>` mock.


fsioni:
Closing this PR because it accidentally included unrelated fork history. I will replace it with a focused branch based on upstream/main.

Issue #584, opened 2026-07-27
Bug: Japanese IME: Enter used to confirm composition also sends the message


TakanoriYanada:
### Environment

- Jean 0.1.68
- macOS Tahoe 26
- Reproduced with:
  - macOS built-in Japanese Input
  - Google Japanese Input

### Description

When typing Japanese, pressing Enter to confirm the current IME composition immediately submits the chat message.

This issue occurs with both the built-in macOS Japanese Input and Google Japanese Input, so it does not appear to be specific to a particular IME.

### Steps to reproduce

1. Open any chat in Jean.
2. Switch the input method to Japanese.
3. Type text that requires IME conversion.
4. Press Enter to confirm the conversion.

### Expected behavior

Pressing Enter should only confirm the current IME composition.

The message should only be sent after pressing Enter again.

### Actual behavior

The Enter key used to confirm the IME composition also submits the message immediately.

### Additional information

- Reproduced 100% of the time.
- This issue only occurs in Jean.
- The same input methods work correctly in:
  - Cursor
  - VS Code
  - ChatGPT Desktop
  - Claude Desktop

This appears to be related to how the chat input handles IME composition events on macOS.

TakanoriYanada:
### Screen recording

The attached screen recording shows the issue.

I type Japanese text and press **Enter** to confirm the IME conversion.  
The conversion is confirmed, but the message is also submitted immediately by the same Enter key.

https://github.com/user-attachments/assets/74a50cf2-b8ac-432f-8103-c5701e5ec364

TakanoriYanada:
I checked the current implementation and the likely location appears to be:

`src/components/chat/ChatInput.tsx`

The `handleKeyDown` handler submits on Enter without checking IME composition state.

A possible fix may be to guard Enter handling with `nativeEvent.isComposing`, `keyCode === 229`, and/or a composition state ref managed by `onCompositionStart` / `onCompositionEnd`.

PR #685, opened 2026-08-15
fix(chat): ignore IME composition Escape in session modal

noih:
## Summary

- Pressing Escape to cancel a CJK IME candidate selection (Chinese/Japanese/Korean) in the chat input also closed the session panel and returned to the project root
- Skip the window-level Escape handler in `SessionChatModal` while a composition is active, reusing the `isImeComposingEvent` guard already applied to Enter in `ChatInput` (#584); covers both `isComposing` and Safari/WKWebView `keyCode 229`
- Add a regression test alongside the existing `SessionChatModal` behavior tests

## Test plan

- [x] `bun run test:run src/components/chat/SessionChatModal.removal-behavior.test.ts`
- [x] `bun run typecheck && bun run lint`
- [x] Manual (macOS, `bun run tauri:dev`): open a session panel, type with a Chinese IME, press Escape while candidates are shown → candidates dismiss, panel stays open; press Escape outside composition → panel closes as before

### DIFF
--- a/src/components/chat/SessionChatModal.removal-behavior.test.ts
+++ b/src/components/chat/SessionChatModal.removal-behavior.test.ts
@@ -185,4 +185,15 @@ describe('SessionChatModal removal behavior', () => {
       /reconnectNativeCliSession\(nativeSession, worktreeId, \{[\s\S]*?openModal: false/
     )
   })
+
+  it('ignores Escape that cancels an IME composition instead of closing the modal', () => {
+    const source = readSource('src/components/chat/SessionChatModal.tsx')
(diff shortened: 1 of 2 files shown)


noih:
**Before** — Escape while the IME candidate bar is showing closes the session panel:

https://github.com/user-attachments/assets/1e86530d-7c04-4893-bd32-f7fbcd24bc10

**After** — Escape only dismisses the candidates; the panel stays open (Escape outside composition still closes it):

https://github.com/user-attachments/assets/68671331-5a9d-45de-916d-1dbf21598b6c

PR #689, opened 2026-08-16
fix(terminal): insert a newline on Shift+Enter

noih:
## Summary

- Make Shift+Enter insert a newline in the embedded terminal instead of submitting: terminals send a bare carriage return for both Enter and Shift+Enter, so Claude Code and friends cannot tell them apart. xterm.js cannot produce the CSI u encoding that would, so inject `\x1b[13;2u` and claim every event of the press, or the renderer adds its own carriage return and the CLI still submits.
- Send it only while the foreground program has shown it reads that encoding — a kitty keyboard protocol push, or focus reporting (`CSI ? 1004 h`, which Claude Code enables). The alternate screen is deliberately not trusted: `vim`, `less` and `fzf` use it without reading CSI u. A plain shell prompt qualifies for neither, so Shift+Enter there still behaves like Enter and never echoes a raw `;2u`.
- Add 15 tests covering the encoding, per-press emission, IME-composition Enter, protocol tracking, and the output chunk-split edge cases.

## Test plan

`bun run check:all` passes. Verified manually on macOS with xterm.js: Shift+Enter inserts a newline in Claude Code, Enter still submits, the zsh prompt is clean, and Ctrl+C/Ctrl+D are unaffected. The `ghostty-web` renderer shares the handler but was not exercised by hand.

### DIFF
--- a/docs/developer/keyboard-shortcuts.md
+++ b/docs/developer/keyboard-shortcuts.md
@@ -378,3 +378,28 @@ fn default_keybindings() -> std::collections::HashMap<String, String> {
 5. **Support migration**: Add to `MIGRATED_KEYBINDINGS` when changing defaults
 6. **Use `mod` prefix**: Allows cross-platform compatibility
 7. **Provide feedback**: Use notifications or UI changes to confirm execution
+
+## Shift+Enter in the embedded terminal
+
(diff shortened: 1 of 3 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #697, opened 2026-08-20
fix(sentry): route Sentry issue lookups to the org's region host

SeriousM:
## Summary

- Selecting a Sentry issue to create a worktree (or load into a session) failed with `404 "Sentry organization or project was not found"` for orgs on a non-default region (e.g. `de.sentry.io`, `eu.sentry.io`).
- Root cause: `sentry_issues.rs` built every API URL from the hardcoded `https://sentry.io` host. Sentry issue IDs are region-scoped, so `GET /issues/{id}` — the call used when creating a worktree from a selected issue — returns 404 on the global host, which the shared helper reported with the misleading org/project message. Listing projects/issues still worked because org/project-level endpoints route globally.
- Fix: resolve the org's region API base URL from `GET /api/0/organizations/` (the same endpoint the project picker uses) and route `list_sentry_issues`, `get_sentry_issue`, and issue-context reads through it. Persist the resolved host as `Project.sentry_base_url`; on resolution failure fall back to `https://sentry.io` **without** caching it, so a transient error never freezes a stale host.
- The cached region host is invalidated when the Sentry auth token or org/project settings change, at both per-project and global scope, so switching to another region's project re-resolves correctly.
- Sentry issue-context removal no longer requires a configured Sentry token, so stale contexts stay removable after the integration is removed.

## Verification

- New unit tests: `resolves_the_organizations_region_base_url`, `falls_back_to_default_without_persisting_region_on_error` (regression for the persistence fix), `persists_a_resolved_region_host` (jean-core).
- Live integration check with a real token (org region `https://de.sentry.io`): `GET /issues/{id}` returned **404 on `sentry.io`** and **200 on the region host**; the latest-event fetch also returned 200.
- Passed: `typecheck`, `lint`, `rust:fmt:check`, `rust:clippy`, `rust:test` (jean-core 1078/0, src-server, src-tauri 14/0). The single `check:all` failure is a pre-existing, unrelated `preferences-prompt-parity.test.ts` (release-notes prompt drift between Rust and TypeScript) that also fails on `main`.

## Breaking Changes

None.

### DIFF
--- a/jean-core/src/auto_fix/scheduler.rs
+++ b/jean-core/src/auto_fix/scheduler.rs
@@ -751,6 +751,7 @@ fn project_from_pending_auto_yolo(entry: &PendingAutoYolo) -> Project {
         sentry_auth_token: None,
         sentry_organization_slug: None,
         sentry_project_slug: None,
+        sentry_base_url: None,
         linked_project_ids: Vec::new(),
         auto_fix_settings: None,
     }
@@ -997,6 +998,7 @@ mod tests {
             sentry_auth_token: None,
             sentry_organization_slug: None,
             sentry_project_slug: None,
+            sentry_base_url: None,
             linked_project_ids: Vec::new(),
             auto_fix_settings,
         }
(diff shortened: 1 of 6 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #702, opened 2026-08-25
fix: prevent projects.json corruption (#679)

### LINKED ISSUE #679: [Bug]: Kernel Crash
### Describe the issue

Version 0.1.72 (sha 32a3b2997d41). Timeline nails down: log timestamps are UTC, boot 15:19 local, crash 14:51 local, last good write 14:42 local.

Paste-ready issue:

# Power loss zeroes `projects.json` → all projects silently disappear, no backup, no recovery

**Version:** 0.1.72 (`32a3b2997d41`, built 2026-08-07)
**OS:** Windows 11 Pro 26200, NTFS
**Severity:** data loss — every project, worktree and session link gone from the UI

## Summary

After an unexpected power loss, `projects.json` was left as **4037 NUL bytes**. The file size was preserved, the contents were not. Jean now fails to parse it on every load, treats the result as "no projects", and shows an empty sidebar. There is no backup file, no recovery path, and no error surfaced in the UI — from the user's side it just looks like every project was deleted.

## Timeline (local time, CEST)

| Time | Event |
|---|---|
| 14:42 | `projects.json` last written (mtime) |
| 14:51 | Unexpected shutdown — Windows event `Kernel-Power 41`, then `6008` |
| 15:19 | Boot |
| 16:14 → | Jean starts, logs a parse failure every 10s, shows zero projects |

## Evidence

`projects.json` is 4037 bytes, entirely zeros:

```
$ od -c projects.json
0000000  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0
*
0007700  \0  \0  \0  \0  \0
0007705
```

Log, repeating every 10 seconds forever:

```
[jean_core::projects::storage][ERROR] Failed to parse projects JSON: expected value at line 1 column 1
[jean_core::auto_fix::scheduler][WARN] Mr. Robot: failed to load projects data: Failed to parse projects data: expected value at line 1 column 1
```

Everything else in `%APPDATA%\com.jean.desktop\` survived intact — `ui-state.json`, `preferences.json`, `.persisted-scope`, `sessions/index/*`, `sessions/data/*`. Only `projects.json` was zeroed, and it was the file most recently written before the crash.

## Root cause (inference — I don't have the source)

A file that comes back the right size but full of NULs after power loss is the standard NTFS signature of a **non-atomic in-place write with no flush to disk**. The filesystem journals the metadata (the new length) but the data blocks were still sitting in the page cache when power dropped, so they read back as zeros.

That points at the write path in `jean_core::projects::storage` doing something equivalent to "open/truncate the real file, serialize into it, close" — no temp file, no `fsync`, no rename.

Three things then compound it:

1. **Single copy.** `projects.json` is the only record of project metadata. Nothing else on disk maps a project UUID to a path or name — `ui-state.json` references project UUIDs but never stores their paths. Lose the one file, lose everything.
2. **Parse failure is treated as "empty".** The loader errors, and the app carries on with an empty project list. A corrupt store and a genuinely empty store are indistinguishable to the user.
3. **The error never reaches the UI.** It only goes to `Jean.log`. The sidebar just renders empty, which reads as "the app deleted my work".

## Impact

- All projects vanish from the UI.
- The `active_session_ids` / `last_opened_per_project` mappings in `ui-state.json` become orphaned — the session data is still on disk but unreachable, since the project UUIDs it points at no longer exist anywhere.
- The only fix available to the user is to delete the corrupt file, re-add every project folder by hand, and accept that chat history is orphaned behind now-dead UUIDs.

## Suggested fixes

**1. Atomic writes with durability, for every JSON store**

```
write payload to projects.json.tmp
fsync(tmp)                 // data actually on platter
rename(tmp -> projects.json)   // atomic swap
fsync(parent directory)    // the rename itself is durable
```

`rename` is atomic on NTFS and POSIX, so a crash leaves either the complete old file or the complete new one — never a zeroed one. Crates like `tempfile::NamedTempFile::persist` or `atomicwrites` do this correctly; the `fsync` before the rename is the part that's easy to omit and is exactly what's missing here.

Worth applying to `ui-state.json`, `preferences.json` and `sessions/index/*.json` too — they survived by luck of timing, not by design.

**2. Keep a rotating backup**

On each successful save, move the previous `projects.json` to `projects.json.bak` (keep 2–3 generations). On a parse failure, fall back to the newest backup that parses.

**3. Never silently downgrade corrupt to empty**

If parsing fails: rename the bad file to `projects.json.corrupt-<timestamp>`, restore from backup if one parses, and show a dialog explaining what happened. Continuing with an empty list — and retrying the same failing read every 10 seconds without ever telling the user — is the part that turns a recoverable file problem into what looks like total data loss.

**4. Cheap rebuild path**

The project paths were still recoverable from `.persisted-scope` (the Tauri fs allowlist) and from `~/.claude/projects/` directory names. A "rescan / rebuild from known paths" action would have made this a 10-second recovery instead of a manual re-add of every project.

## Reproduction

1. Open Jean with several projects configured.
2. Trigger a save of `projects.json` (add or modify a project).
3. Cut power within a few minutes — hard power-off, not a clean shutdown, so the page cache is never flushed.
4. Boot and start Jean: `projects.json` is NUL-filled and the project list is empty.

## Workaround for anyone hitting this

Quit Jean, rename `%APPDATA%\com.jean.desktop\projects.json` to `.corrupt`, restart, re-add the project folders. To recover orphaned session history, note the project UUIDs in `ui-state.json` (`project_access_timestamps`) before re-adding, then edit the new `projects.json` to restore those UUIDs.

### Replication

_No response_

### How are you running Jean?

App

### Operating system

Windows 11

### Jean version

v0.1.72

### How are you interacting with Jean?

Via the app

manstfu:
## Summary

- Make `projects.json` writes durable with flushed unique temp files and atomic replacement.
- Add rotating backups and automatic recovery/quarantine for corrupt project data.
- Surface project-load failures instead of presenting an empty project list.
- Document the crash-safe persistence pattern.

Fixes #679

## Validation

- Full `jean-core` Rust test suite: 1018 passed.
- Rust clippy passed for `jean-core`, `src-server`, and `src-tauri`.
- TypeScript, changed-file ESLint, and Prettier checks passed.
- Targeted frontend tests passed; repository-wide checks still encounter pre-existing unrelated lint/test failures.

### DIFF
--- a/docs/developer/data-persistence.md
+++ b/docs/developer/data-persistence.md
@@ -130,6 +130,25 @@ async fn save_preferences(app: AppHandle, preferences: AppPreferences) -> Result
 
 **Note:** Using UUID in temp filenames (instead of just `.tmp`) prevents conflicts when multiple writes happen concurrently.
 
+### Durable Writes and Recovery for Critical Data
+
+For user data where losing the file can make the application appear empty, an atomic
+rename alone is not sufficient: the temporary file must be flushed before it becomes
+the live file. The project registry (`projects.json`) uses the stronger pattern in
+`jean-core/src/projects/storage.rs`:
+
+- Write to a unique sibling temporary file with `create_new`.
+- Call `sync_all` before atomically replacing the destination; also sync the parent
+  directory on Unix and use write-through replacement on Windows.
+- Keep three rotating backups (`projects.json.bak`, `.bak.1`, and `.bak.2`).
+- If the primary JSON is invalid, validate backups and leftover temporary files,
+  preserve the corrupt primary, and restore the first valid candidate.
+- If recovery is impossible, return an error. Never convert a failed read into an
+  empty collection in the UI or an API response.
+
+Use this pattern for other critical persisted data instead of silently falling back
+to defaults after a parse failure.
+
 ### Loading with Defaults
 
 ```rust
@@ -562,7 +581,7 @@ function migrateKeybindings(
 
 ## Best Practices
 
-1. **Use atomic writes**: Always write to temp file then rename
+1. **Use durable atomic writes**: Flush the temp file before replacing the destination, and sync the parent directory where supported
 2. **Validate inputs**: Check filenames and data before writing
 3. **Handle defaults**: Provide sensible defaults when files don't exist
 4. **Log operations**: Log all file operations for debugging
--- a/jean-core/Cargo.toml
+++ b/jean-core/Cargo.toml
@@ -38,7 +38,7 @@ libc = "0.2"
 tokio-tungstenite = "0.28"
 
 [target.'cfg(windows)'.dependencies]
-windows-sys = { version = "0.59", features = ["Win32_System_Threading", "Win32_Foundation"] }
+windows-sys = { version = "0.59", features = ["Win32_System_Threading", "Win32_Foundation", "Win32_Storage_FileSystem"] }
 
 [dev-dependencies]
 tempfile = "3.23.0"
--- a/jean-core/src/http_server/server.rs
+++ b/jean-core/src/http_server/server.rs
@@ -547,6 +547,7 @@ async fn init_handler(
     response["serverPlatform"] = Value::String(crate::server_platform_name().to_string());
     response["nativeOpenAllowed"] = Value::Bool(crate::platform::native_open_allowed());
 
+    let projects_load_failed = projects_result.is_err();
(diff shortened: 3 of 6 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #705, opened 2026-08-25
perf: reduce CPU and memory across projects

manstfu:
## Summary

- Reduce project/session retention by loading full worktree and session data only for the active canvas project.
- Replace broad unread scans with count-only/index-backed queries.
- Coalesce and bound Git-status work, batch cache writes, and limit background sweeps.
- Bound detached terminal renderers/scrollback and evict long-idle renderers while preserving live PTYs.
- Clean session/worktree keyed UI state when sessions close.
- Reduce canvas rendering/query overhead with cached card derivation, viewport-gated status subscriptions, search normalization, and stable query definitions.

## Validation

- `bun run typecheck` ?
- `bun run lint` ?
- Focused frontend tests ?
- `git diff --check` ?

`bun run check:all` currently stops at pre-existing Rust formatting drift in unrelated files. The full test run also has two known baseline failures unrelated to this change.


andrasbacsai:
Thank you for the PR! 💜

PR #726, opened 2026-09-06
fix(chat): restore cancelled runs to input when no output streamed

### LINKED ISSUE #725: [Bug]: canceling a message causes it to disappear, requiring the user to retype it.
### Describe the issue

This is important because canceling a message should not result in the user losing the work they have already typed. If a message disappears after cancellation, users have to retype it, which is frustrating, wastes time, and increases the risk of losing important information.

Preserving the message content after cancellation provides a safer and more predictable experience, especially for longer or more detailed messages.

### Replication

https://github.com/user-attachments/assets/e2574c3e-1045-487d-b059-846c4cc34cdf

### How are you running Jean?

App

### Operating system

_No response_

### Jean version

0.1.73

### How are you interacting with Jean?

Via the app

Selnapenek:
- Include cancelled runs without assistant content in message restoration logic, since the backend omits such runs from chat history. 
- This ensures the cancelled prompt text is restored to the input, matching backend state.

fixes: https://github.com/coollabsio/jean/issues/725

https://github.com/user-attachments/assets/6a103841-87e7-4bc9-95c1-5794cb2c1eb2

### DIFF
--- a/src/components/chat/hooks/useStreamingEvents.ts
+++ b/src/components/chat/hooks/useStreamingEvents.ts
@@ -1929,12 +1929,14 @@ export default function useStreamingEvents({
         // Clear compacting state (safety net)
         useChatStore.getState().setCompacting(session_id, false)
 
-        // Restore message to input ONLY when the prompt never started
-        // (backend undo_send=true: process not registered / pending cancel).
-        // If the prompt is already running, do not restore even when no
-        // assistant content has streamed yet — cancel of a live run leaves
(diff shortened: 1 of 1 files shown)


Selnapenek:
the issue fixed

andrasbacsai:
Closed as a fork PR (maintainer edits were off, so we could not push follow-up commits here). The same fix, plus tests and the hydrate/duplicate-draft handling, landed on `main` via https://github.com/coollabsio/jean/pull/731.

PR #754, opened 2026-09-17
fix(test): remove duplicate sentry_base_url initialiser

Dredok:
## Summary

`jean-core`'s test target currently fails to compile on `main`, so `cargo test` / `bun run rust:test` cannot run at all.

`failed_project_avatar_copy_keeps_the_previous_file` (`jean-core/src/projects/commands.rs`) initialises `sentry_base_url` twice in the same struct literal. Duplicate field initialisers are a hard error (E0062), not a warning:

```
error[E0062]: field `sentry_base_url` specified more than once
     --> src/projects/commands.rs:15390:13
```

Release builds are unaffected because the code is inside `#[cfg(test)]`, which is presumably why this went unnoticed.

This removes the second initialiser. Both were `None`, so there is no behavioural change to the test.

## Test plan

- [x] `cargo test --lib` on `jean-core` now builds and passes: **1216 passed, 0 failed, 1 ignored**
- [x] Before this change the same command aborts with `could not compile jean-core (lib test)`
- [x] `cargo fmt --check` clean

Built and tested in the `rust:1-bookworm` image used by `Dockerfile.server`.


Dredok:
Closing: upstream fixed this independently. `jean-core/src/projects/commands.rs` on current `main` no longer has the duplicate `sentry_base_url` initialiser, and `cargo test --lib` builds and passes without this patch. Thanks!

